PR #2, opened 2017-08-30
python2 fixes

yhenon:
fixes for python2 compatibility:
- updated `super()` to python2/3 compatible syntax
- removed non-unicode character


hgaiser:
> removed non-unicode character
I was a bit surprised by this, but apparently that character is a weird non-ASCII `~`. 
I changed it to a normal `~`. Other than that, looks good! Thanks!

PR #3, opened 2017-09-05
Loss calc

yhenon:
Divide by `1` when there is no assigned boxes, instead of dividing by `keras.backend.epsilon()`. Stabilises training significantly.


hgaiser:
Looks good, thanks!

PR #5, opened 2017-09-20
python2 support

yhenon:
Just to maintain python 2 support


hgaiser:
Thanks! I replaced the import with changing the divisions themselves. Since they are only in two places I felt that made more sense.

PR #7, opened 2017-09-22
add AR preserving resize

yhenon:
Images are resized according to faster r-cnn technique. Maybe there should also be an option for fixes resizing or no scaling?


hgaiser:
Moved to `image.py` so that it can be used in other generators too.

Thanks!

PR #8, opened 2017-09-26
updated for new losses

yhenon:
Updated to support pascal voc on the refactored loss. Also includes small fixes for py2 support.


hgaiser:
Looks good, thanks!

PR #13, opened 2017-10-04
Change kernel size in pyramic convolutions.

enricoliscio:
Modified the convolutional filters dimensions when upsampling and adding for climbing the pyramid features, as proposed in the paper.
When moving from P5 to P4, there is upsampling of P5 and addition of a 1x1 convolution of C4, not a 3x3 convolution.


hgaiser:
Nice catch!

PR #17, opened 2017-10-10
Refactor how anchors are computed.

hgaiser:
This PR does 3 things:

- Refactor the way anchors are generated
- Changes the `Anchor` layer such that it uses the same (base) anchor generation code as the data generators.
- Corrects (base) anchor generation such that they are centered around the point `(0, 0)`.


enricoliscio:
Good!

PR #18, opened 2017-10-10
Separate classification loss in Pascal training set.

enricoliscio:
I had forgotten to separate regression and classification in the Pascal preprocessing. Now also the Pascal training can be run.


hgaiser:
Looks good!

PR #26, opened 2017-10-23
Create LICENSE

hgaiser:


de-vri-es:
The license file itself is fine of course (although it uses different brackets in the appendix from the plain text version at apache.org [1]  for some reason, its identical otherwise).

But as the appendix says, all copyrighted files should also include a copyright notice. The appendix has more details regarding the form of the copyright notice.

[1] https://www.apache.org/licenses/LICENSE-2.0.txt

hgaiser:
Added the header to the python files. All good ?

de-vri-es:
Looks good. Interesting choice to use string literals instead of comments, but that doesn't really matter I think.

PR #27, opened 2017-10-25
Sigmoid activation

hgaiser:
This PR changes the classification activation function to a Sigmoid activation. This has the following consequences:

- There is no longer a `__background__` class. Background detections are represented by having all class scores close to 0.
- Multi-class labels are possible using sigmoid, not using softmax.

The paper also describes using a sigmoid and preliminary results show a slight improvement over using the softmax activation (though should be taken with a grain of salt).


enricoliscio:
Seems good to me.

mihaimorariu:
Why are you using one-hot encoding in `anchor_targets`, for generating `labels`? Seems good to me as well.

hgaiser:
The label targets are now one-hot because the background class is no longer a label. Meaning background labels are represented by having all values in the one-hot vector set to 0.

PR #28, opened 2017-10-25
Unit tests

hgaiser:
This PR adds unit test for the common backend and the miscellaneous layers. Currently some tests are expected to fail (and marked as such) because `batch_size > 1` is not yet supported.


mihaimorariu:
Seems good to me.

PR #29, opened 2017-10-31
Smooth L1 loss for bounding box regression

hgaiser:
This PR implements smooth L1 loss for bounding box regression, as per the paper.


enricoliscio:
Seems good.

PR #30, opened 2017-10-31
Add COCO evaluation tool.

hgaiser:
Add tool for evaluating COCO models.


enricoliscio:
Apart from the COCO specific part that I don't really know, it seems good.

PR #36, opened 2017-11-02
Keras 2.0.9

hgaiser:
This PR makes the necessary adjustments for Keras 2.0.9. Most noticeably:

- The interface for dataset generators changed slightly, introducing a `_get_batches_of_transformed_samples` function which computes the input/output batch(es) given a list of indices.
- `TensorReshape` can be removed since https://github.com/fchollet/keras/pull/7960 was accepted (which was the reason a custom `TensorReshape` layer was added to `keras-retinanet` to begin with).


enricoliscio:
Seems good. If Mr. Keras accepted your pull reuqest, it must be working.

PR #41, opened 2017-11-07
fix CocoIterator being loaded even though we're not using COCO datase…

sorinpanduru:


hgaiser:
I agree with the change, but this change will break the [examples](https://github.com/delftrobotics/keras-retinanet/blob/master/examples). Also to be consistent, can you also remove the VocIterator and make the appropriate changes in the examples?

sorinpanduru:
Hi!
I updated the examples scripts too.

hgaiser:
Thanks, almost there. It's used in the jupyter notebooks and the COCO evaluation script as well.

sorinpanduru:
Hello,
the notebooks and the coco eval scripts use a FQN when instantiating the CocoIterator class... there's no need to do the imports there too.

hgaiser:
Eventhough they use a FQN, if you don't import it in the `__init__.py` it will not automatically import that submodule. Try it out, I thought it would work too but when I tested it, it doesn't work.

hgaiser:
If you import it as `from keras_retinanet.preprocessing.coco import CocoIterator`, then it also needs to be referenced as `CocoIterator`. Currently it is being referenced as `keras_retinanet.preprocessing.coco.CocoIterator`.

hgaiser:
Awesome, looks good.

PR #51, opened 2017-11-10
CSV based generator

yhenon:
Adds a new generator, based on ingesting a simple csv file. Useful feature for people to train on their own dataset!

### DIFF
--- a/examples/train_csv.py
+++ b/examples/train_csv.py
@@ -1,4 +1,6 @@
 """
+Copyright 2017-2018 yhenon (https://github.com/yhenon/)
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -39,8 +41,9 @@ def create_model(num_classes, weights='imagenet'):
 def parse_args():
     parser = argparse.ArgumentParser(description='Simple training script for object detection from a CSV file.')
     parser.add_argument('csv_path', help='Path to CSV file')
+    parser.add_argument('--classes', help='Path to a CSV file containing class label mapping.')
     parser.add_argument('--weights', help='Weights to use for initialization (defaults to ImageNet).', default='imagenet')
-    parser.add_argument('--batch-size', help='Size of the batches.', default=1, type=int)
+    parser.add_argument('--batch-size', help='Size of the batches.', default=2, type=int)
     parser.add_argument('--gpu', help='Id of the GPU to use (as reported by nvidia-smi).')
 
     return parser.parse_args()
@@ -60,19 +63,24 @@ if __name__ == '__main__':
     )
     test_image_data_generator = keras.preprocessing.image.ImageDataGenerator()
 
+
     # create a generator for training data
     train_generator = CSVGenerator(
-        args.csv_path,
-        'train',
-        train_image_data_generator,
+        csv_data_file=args.csv_path,
+        set_name='train',
+        csv_class_file=args.classes,
+        image_data_generator=train_image_data_generator,
         batch_size=args.batch_size
     )
+    while True:
+        X, Y = train_generator.next()
 
     # create a generator for testing data
     test_generator = CSVGenerator(
-        args.csv_path,
-        'val',
-        test_image_data_generator,
+        csv_data_file=args.csv_path,
+        set_name='val',
+        csv_class_file=args.classes,
+        image_data_generator=test_image_data_generator,
         batch_size=args.batch_size
     )
 
--- a/examples/train_csv.py
+++ b/examples/train_csv.py
@@ -1,0 +1,116 @@
+"""
+Copyright 2017-2018 Fizyr (https://fizyr.com)
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import argparse
+import os
+
+import keras
+import keras.preprocessing.image
+
+import tensorflow as tf
+
+from keras_retinanet.models import ResNet50RetinaNet
+from keras_retinanet.preprocessing.csv_generator import CSVGenerator
+import keras_retinanet
+
+
+def get_session():
+    config = tf.ConfigProto()
+    config.gpu_options.allow_growth = True
+    return tf.Session(config=config)
+
+
+def create_model(num_classes, weights='imagenet'):
+    image = keras.layers.Input((None, None, 3))
+    return ResNet50RetinaNet(image, num_classes=num_classes, weights=weights)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Simple training script for object detection from a CSV file.')
+    parser.add_argument('csv_path', help='Path to CSV file')
+    parser.add_argument('--weights', help='Weights to use for initialization (defaults to ImageNet).', default='imagenet')
+    parser.add_argument('--batch-size', help='Size of the batches.', default=1, type=int)
+    parser.add_argument('--gpu', help='Id of the GPU to use (as reported by nvidia-smi).')
+
+    return parser.parse_args()
+
+if __name__ == '__main__':
+    # parse arguments
+    args = parse_args()
+
+    # optionally choose specific GPU
+    if args.gpu:
+        os.environ['CUDA_VISIBLE_DEVICES'] = args.gpu
+    keras.backend.tensorflow_backend.set_session(get_session())
+
+    # create image data generator objects
+    train_image_data_generator = keras.preprocessing.image.ImageDataGenerator(
+        horizontal_flip=True,
+    )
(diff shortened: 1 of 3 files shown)


hgaiser:
Looks good, but I'm a bit concerned about how labels are created. Seems like it now depends on the order as they appear in the csv file? Perhaps it's better to sort them by name and then assign labels?

yhenon:
I like that suggestion. Are you also considering a strategy to serialise
the labels to keep them consistent between train & test time?

hgaiser:
Maybe it would make more sense to have the first row describe the classes? That way it can indeed be consistent. The method I proposed will also fail if there is not at least 1 annotation per object class in your CSV file.

de-vri-es:
It's not exactly in the spirit of csv to have a separate list on the first line. You would normally only use it for a header for the remaining data, or not have it at all. But it seems the only other viable option would be a separate file. I'm not sure what is worse :)

de-vri-es:
After thinking a bit more, I think that it does make sense to keep the class names in a separate file.

One reason is that the class name/number mapping is still useful without annotation. For example, a program running the network could use it to show human readable names.

Another reason is that it makes it really easy to combine and split CSV files with annotations. Tools like `cat` and `cut` will work perfectly. If we'd put class names on the first lines you would always need to treat the first line specially.

Note that it would be possible to add an option for automatically determining a mapping as an alternative to loading the mapping from a file.

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -0,0 +1,117 @@
+"""
+Copyright 2017-2018 Fizyr (https://fizyr.com)
```
de-vri-es: Thanks for the contribution, I really like the simplicity and re-usability of this generator.

We can't claim copyright for this though. In general you can not give away copyright without signing a legal document. So you should put your own copyright notice here.

If you developed this as part of your job (as employee) you should credit your employer. Otherwise, you can credit yourself (or a pseudonym if you don't want personal details in the file).

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -1,6 +1,4 @@
 """
-Copyright 2017-2018 Fizyr (https://fizyr.com)
```
de-vri-es: Is there anything we can add as a copyright statement? Your real name, or otherwise your github name perhaps?

de-vri-es: Or should it be Netra Inc. ?

examples/train_csv.py
```diff
@@ -39,8 +41,9 @@ def create_model(num_classes, weights='imagenet'):
 def parse_args():
     parser = argparse.ArgumentParser(description='Simple training script for object detection from a CSV file.')
     parser.add_argument('csv_path', help='Path to CSV file')
+    parser.add_argument('--classes', help='Path to a CSV file containing class label mapping.')
     parser.add_argument('--weights', help='Weights to use for initialization (defaults to ImageNet).', default='imagenet')
-    parser.add_argument('--batch-size', help='Size of the batches.', default=1, type=int)
+    parser.add_argument('--batch-size', help='Size of the batches.', default=2, type=int)
```
hgaiser: Why change the default to 2?

examples/train_csv.py
```diff
@@ -60,19 +63,24 @@ def parse_args():
...
+    while True:
+        X, Y = train_generator.next()
 
     # create a generator for testing data
     test_generator = CSVGenerator(
-        args.csv_path,
-        'val',
-        test_image_data_generator,
+        csv_data_file=args.csv_path,
+        set_name='val',
```
hgaiser: Maybe make this an argument, defaulting to `val`.

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -26,28 +28,46 @@
...
-        # csv with img_filepath, class, x1, y1, x2, y2, dataset
-        with open(csv_file, 'rb') as f_in:
+
+        if csv_class_file is not None:
+            #parse the provided class file
+            autogen_classes = False
+            self.classes = {}
+            with open(csv_class_file, 'rb') as f_class_in:
+                csvreader = csv.reader(f_class_in, delimiter=',')
+                for row in csvreader:
```
hgaiser: Maybe change this to `for classname, class_id in csvreader:`

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -26,28 +28,46 @@
...
+                    classname, class_id = row
+                    self.classes[classname] = int(class_id)
+        else:
+            autogen_classes = True
+            self.classes              = []
```
hgaiser: Alignment

yhenon:
Alright I'm fixing this up:
- there is now an optional `classname` argument in `examples/train_csv.py`. This can be used to pass a CSV file containing classname->id mappings. The parser will then respect this mapping, and throw an error if a class present in the dataset is missing in the mapping. For example the user could pass a CSV containing:
```
cat,0
dog,99
```
This would then create a 100 class model, with the required mappings.
- by default, the mappings are auto-generated by parsing the CSV. The classes are now sorted before being assigned an ID
- the copyright notice has been added
- some updates to the README to explain the use of the csv generator

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -57,6 +77,13 @@ def __init__(
                     self.image_data[img_filepath] = []
                 self.image_data[img_filepath].append({'x1':int(x1), 'x2':int(x2), 'y1': int(y1), 'y2':int(y2), 'class':classname})
 
+        # sort the list of classes and convert to a dict
+
+        if autogen_classes:
+            self.classes = dict([(classname,ix) for ix, classname in enumerate(sorted(self.classes))])
+
+        print(self.classes)
```
hgaiser: I wouldn't print here.

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -105,10 +132,10 @@ def load_annotations(self, image_index):
             box[0, 4] = self.name_to_label(class_name)
 
             bndbox = [annot['x1'], annot['y1'], annot['x2'], annot['y2']]
-            box[0, 0] = float(bndbox[0]) - 1
-            box[0, 1] = float(bndbox[1]) - 1
-            box[0, 2] = float(bndbox[2]) - 1
-            box[0, 3] = float(bndbox[3]) - 1
+            box[0, 0] = float(bndbox[0])-1
```
hgaiser: Why remove the spaces?

yhenon:
> I was thinking if it makes more sense to have multiple files per set. So you'd have a train.txt CSV file for your training dataset and a test.txt file for your test dataset, etc. What do you think?

I agree. The user should now pass a CSV file for training, and, optionally a CSV file for validation.

In addition, passing a CSV file with classes is now required. This avoids difficulties associated with automatically inferring the classes for both the train and val file, especially if some classes are missing in one. The class file can be reused at inference time.

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -0,0 +1,117 @@
...
+        return cv2.imread(path)
+
+    def load_annotations(self, image_index):
+        boxes = np.zeros((0, 5))
+
+        path  = self.image_names[image_index]
+        image = Image.open(path)
+
+        width = float(image.width)
+        height = float(image.height)
```
mateuszbuda: unused variables

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -0,0 +1,117 @@
...
+
+            class_name = annot['class']
+
+            box[0, 4] = self.name_to_label(class_name)
+
+            bndbox = [annot['x1'], annot['y1'], annot['x2'], annot['y2']]
+            box[0, 0] = float(bndbox[0]) - 1
+            box[0, 1] = float(bndbox[1]) - 1
+            box[0, 2] = float(bndbox[2]) - 1
+            box[0, 3] = float(bndbox[3]) - 1
```
mateuszbuda: @yhenon Does this mean that you expect indexing of pixels in csv file to be 1 based (starting from 1) and here you transform it to 0 based, i.e. the first pixel has index 0?

keras_retinanet/preprocessing/csv_generator.py
yhenon: unless I'm confused, you are looking at an outdated commit

keras_retinanet/preprocessing/csv_generator.py
yhenon: this matches the behaviour in keras_retinanet/preprocessing/pascal_voc.py

I guess we can either: (a) define the csv format to index starting at 0 and remove the `-1` here, or (b) define the csv format to index starting at 1 and keep the `-1` here.

de-vri-es: Personally I feel we shouldn't copy weird choices made by pascal VOC. One-based indexing is odd. I'm in favor of dropping the -1 and defining the CSV format using zero based indices.

mateuszbuda: As long as it's documented somewhere, e.g. in readme, both options are good for me.

yhenon: fixed as per @de-vri-es comment which I agree with

hgaiser:
Looks good to me. If @de-vri-es agrees I'd like  to merge this. I'm thinking to squash the commits to one commit too.

Issue #47, opened 2017-11-09
 raise StopIteration(e) StopIteration


zccoder:
The problem occur like that:
```
None
Epoch 1/50
Traceback (most recent call last):
  File "examples/train_pascal.py", line 105, in <module>
    keras.callbacks.ReduceLROnPlateau(monitor='val_loss', factor=0.1, patience=10, verbose=1, mode='auto', epsilon=0.0001, cooldown=0, min_lr=0),
  File "/usr/local/lib/python2.7/dist-packages/keras/legacy/interfaces.py", line 87, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keras/engine/training.py", line 2046, in fit_generator
    generator_output = next(output_generator)
  File "/usr/local/lib/python2.7/dist-packages/keras/utils/data_utils.py", line 518, in get
    raise StopIteration(e)
StopIteration

```
It seems that fit_generator has problem. More detailedly, it may be "PascalVocIterator" problem.
I do not know whether it is the version or the code problem?

hgaiser:
Is there no other error beyond this? Usually it also shows where in the generator the error occurred.

dasolyn:
I have a similar issue, too.

```
Traceback (most recent call last):
  File "train_pascal.py", line 109, in <module>
    keras.callbacks.ReduceLROnPlateau(monitor='val_loss', factor=0.1, patience=10, verbose=1, mode='auto', epsilon=0.0001, cooldown=0, min_lr=0),
  File "/usr/local/lib/python2.7/dist-packages/keras/legacy/interfaces.py", line 87, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keras/engine/training.py", line 2046, in fit_generator
    generator_output = next(output_generator)
  File "/usr/local/lib/python2.7/dist-packages/keras/utils/data_utils.py", line 518, in get
    raise StopIteration(e)
StopIteration: 'NoneType' object has no attribute 'text'
```

I edited some code in train_pascal.py because I want to use my own dataset. But I didn't edit pascal_voc.py

dasolyn:
I made a xml annotations file myself so it doesn't contain `difficult`, `truncated` tags.
I added it and now it works well.

zccoder:
@hgaiser 
[...]

PR #54, opened 2017-11-13
Improve error output of PascalVocGenerator.

de-vri-es:
This PR should improve the error messages when the annotations XML files do not match the expected format.

Note that I didn't test yet, but I wanted to let you know I'm working on this.

Should help prevent issues like #47

Note: This is a PR to merge into the `generator-refactor` branch

### DIFF
--- a/keras_retinanet/preprocessing/pascal_voc.py
+++ b/keras_retinanet/preprocessing/pascal_voc.py
@@ -19,10 +19,14 @@ from keras_retinanet.preprocessing.generator import Generator
 import cv2
 import os
 import numpy as np
+from future.utils import raise_from
 from PIL import Image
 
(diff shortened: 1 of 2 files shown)


de-vri-es:
TODO: Raising a new error in an exception handler is not nice. It causes python to say an error occurred during an error handler.

de-vri-es:
Using `raise_from()` (from `future.utils`) fixes the problem of doubly printed exceptions. `raise_from()` is a python2/3 compatible version of `raise ... from ...` (python2/3 syntax diverges here pretty badly). 

@hgaiser: I think it should be suitable for merging now.

hgaiser:
LGTM, thanks!

PR #59, opened 2017-11-14
Check keras version in examples.

de-vri-es:
This should prevent issues on the tracker for people using the wrong keras version.


hgaiser:
LGTM, thanks!

Issue #64, opened 2017-11-16
OverflowError in Windows


dasolyn:
Line 89 in /keras_retinanet/preprocessing/pascal_voc.py,

```python
seed = np.uint32(time.time() * 1000)
```

This line always raise OverflowError in Windows but linux don't.

erhwenkuo:
According to [What is bit size of lon on 64-bit Window?](https://stackoverflow.com/questions/384502/what-is-the-bit-size-of-long-on-64-bit-windows) , apparently the issue is caused by the data type definition between Windows and *nix-like system.

After checking the logic, I think the purpose of below code block is to get a random seed which would keep `keras.preprocessing.image.Iterator` to have reproduciable iteration patten.

Code before change:
`
seed = np.uint32(time.time() * 1000)
`

I think it should be ok to remodify the code block like below:
`
seed = np.uint32(time.time())
`

hgaiser:
Resolved in https://github.com/fizyr/keras-retinanet/pull/48.

PR #65, opened 2017-11-16
Fix: OverflowError in Windows #64

erhwenkuo:
Fix data type bit size different between Windows and *nix operation system, which cause "OverflowError".
Base on the explaination of [What is the bit size of long on 64-bit Windows?](https://stackoverflow.com/questions/384502/what-is-the-bit-size-of-long-on-64-bit-windows).

### DIFF
--- a/keras_retinanet/preprocessing/coco.py
+++ b/keras_retinanet/preprocessing/coco.py
@@ -27,6 +27,7 @@ import cv2
 import os
 import numpy as np
 import time
+import platform
 
 from pycocotools.coco import COCO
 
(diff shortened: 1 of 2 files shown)


hgaiser:
Oh Windows...

Anyway I see what the issue is you're trying to fix. The code was intended to give a a different seed everytime you run it. The current time is a good seed in this case, but casting `time.time()` to int directly gives you only the number of seconds. This means if you create 2 generators during the same second they do the exact same things. I propose to change it to : `(time.time() % 1) * 1000`. That way you get the number of milliseconds, which should provide different seeds everytime.

In addition, I'm in the process of merging https://github.com/fizyr/keras-retinanet/pull/48. Merging this PR with master and then merging that PR will cause conflicts. If you don't mind I'd prefer to close this issue, fix it as proposed in that PR and then merge that. Sounds okay?

erhwenkuo:
Sounds good to me!

PR #72, opened 2017-11-19
use os.path.join for specifying paths

yhenon:
Use `os.path.join` instead of '/' for our friends on Windows.


hgaiser:
Looks good, thanks!

PR #75, opened 2017-11-19
Blacklist keras 2.1.0 and 2.1.1 in utils/keras_version.

de-vri-es:
Alternatively we could just bump the minimum required version to 2.1.1, but then we're being unnecessarily restrictive.

### DIFF
--- a/keras_retinanet/utils/keras_version.py
+++ b/keras_retinanet/utils/keras_version.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 import keras
 import sys
 
-minimum_keras_version = 2, 0, 9
+minimum_keras_version = 2, 1, 2
 
 
 def keras_version():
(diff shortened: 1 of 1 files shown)


hgaiser:
Actually, 2.1.1 also doesn't work. Only `master` and `2.0.9` work at the moment..

de-vri-es:
Ah, I added 2.1.1 to the blacklist. Problem there is that the keras master branch reports it's version as 2.1.1. So the check would also complain about the master branch.

We'll have to hold off merging this then until 2.1.2 :[

hgaiser:
Yep I'm afraid so too. Wouldn't it make more sense for Keras to bump their version after they released, so that `master` has a higher version? Hmm...

de-vri-es:
Yeah, although not to exactly 2.1.2. `2.1.2.dirty` or something similar would be best.

hgaiser:
Keras 2.1.2 is released by the way. In my opinion we should enforce 2.1.2, since 2.0.9 also has some small (non-breaking) bugs.

de-vri-es:
Removed the version blacklist stuff and bumped the minimum version instead. It can be re-added if we ever need it.

I did keep the updated assertion check, since it matches the error message better.

hgaiser:
LGTM

PR #79, opened 2017-11-20
Fix loss calculation when batch_size>1

yhenon:
This PR fixes the loss calculation when batch_size>1. For each image in the minibatch, the loss is normalised by the number of positive anchors. So for a minibatch consisting of 2 images, which have unnormalised loss L1 and L2, and number of positive anchors N1 and N2, the loss is:
```
loss = L1/N1 + L2/N2
```
The current implementation uses (incorrectly):
```
loss = (L1+L2)/(N1+N2)
```

For testing, we can verify that the loss is the same when calculated on individual images vs on a minibatch of images, like so:
```
    import numpy as np

    X, Y = train_generator.next()
    loss_total, loss_regr, loss_class = model.test_on_batch(X, Y)
    loss_total_b =[]
    loss_regr_b =[]
    loss_class_b = []

    for i in range(X.shape[0]):
        L = model.test_on_batch(X[i:i+1, ...], [Y[0][i:i+1, ...], Y[1][i:i+1,...]])
        loss_total_b.append(L[0])
        loss_regr_b.append(L[1])
        loss_class_b.append(L[2])

    np.testing.assert_almost_equal(loss_total, np.mean(loss_total_b), decimal=5)
    np.testing.assert_almost_equal(loss_regr, np.mean(loss_regr_b), decimal=5)
    np.testing.assert_almost_equal(loss_class, np.mean(loss_class_b), decimal=5)
    
    print('passed')
```
The current implementation fails:
```
AssertionError: 
Arrays are not almost equal to 5 decimals
 ACTUAL: 1.6648083
 DESIRED: 1.6746489
```
This PR fixes the issue.

### DIFF
--- a/keras_retinanet/losses.py
+++ b/keras_retinanet/losses.py
@@ -23,17 +23,16 @@ def focal(alpha=0.25, gamma=2.0):
 
         # compute the divisor: for each image in the batch, we want the number of positive anchors
 
-        # override the -1 labels, since we treat -1/0 the same way for determining the divisor
-        divisor   = keras_retinanet.backend.where(y_true < 0, keras.backend.zeros_like(y_true), y_true)
-        divisor   = keras.backend.max(divisor, axis=2)
+        # override the -1 labels, since we treat values -1 and 0 the same way for determining the divisor
+        divisor   = keras_retinanet.backend.where(keras.backend.less_equal(y_true, 0), keras.backend.zeros_like(y_true), y_true)
+        divisor   = keras.backend.max(divisor, axis=2, keepdims=True)
         divisor   = keras.backend.cast(divisor, keras.backend.floatx())
+
         # compute the number of positive anchors
-        divisor   = keras.backend.sum(divisor, 1)
+        divisor   = keras.backend.sum(divisor, axis=1, keepdims=True)
+
         #  ensure we do not divide by 0
         divisor   = keras.backend.maximum(1.0, divisor)
-        # pad the tensor to have shape (batch_size, 1, 1) for future division
-        divisor   = keras.backend.expand_dims(divisor, axis=1)
-        divisor   = keras.backend.expand_dims(divisor, axis=2)
 
         labels         = y_true
         classification = y_pred
@@ -46,7 +45,8 @@ def focal(alpha=0.25, gamma=2.0):
 
         cls_loss = focal_weight * keras.backend.binary_crossentropy(labels, classification)
 
-        cls_loss = cls_loss/divisor
+        # normalise by the number of positive anchors for each entry in the minibatch
+        cls_loss = cls_loss / divisor
 
         # filter out "ignore" anchors
         anchor_state   = keras.backend.max(labels, axis=2)  # -1 for ignore, 0 for background, 1 for object
@@ -54,9 +54,11 @@ def focal(alpha=0.25, gamma=2.0):
 
         cls_loss = keras_retinanet.backend.gather_nd(cls_loss, indices)
 
-        return keras.backend.sum(cls_loss)/keras.backend.cast(keras.backend.shape(y_true)[0], keras.backend.floatx())
+        # divide by the size of the minibatch
+        return keras.backend.sum(cls_loss) / keras.backend.cast(keras.backend.shape(y_true)[0], keras.backend.floatx())
 
     return _focal
+
 
 def smooth_l1(sigma=3.0):
     sigma_squared = sigma ** 2
@@ -70,11 +72,10 @@ def smooth_l1(sigma=3.0):
 
         # compute the divisor: for each image in the batch, we want the number of positive anchors
         divisor = keras_retinanet.backend.where(keras.backend.equal(anchor_state, 1), anchor_state, keras.backend.zeros_like(anchor_state))
-        divisor   = keras.backend.sum(divisor, axis=1)
+        divisor   = keras.backend.sum(divisor, axis=1, keepdims=True)
         divisor   = keras.backend.maximum(1.0, divisor)
 
         # pad the tensor to have shape (batch_size, 1, 1) for future division
-        divisor   = keras.backend.expand_dims(divisor, axis=1)
         divisor   = keras.backend.expand_dims(divisor, axis=2)
 
         # compute smooth L1 loss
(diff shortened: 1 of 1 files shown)


keras_retinanet/losses.py
```diff
@@ -20,48 +20,62 @@
...
-        # filter out "ignore" anchors
-        anchor_state   = keras.backend.max(labels, axis=1)  # -1 for ignore, 0 for background, 1 for object
-        indices        = keras_retinanet.backend.where(keras.backend.not_equal(anchor_state, -1))
-        classification = keras_retinanet.backend.gather_nd(classification, indices)
-        labels         = keras_retinanet.backend.gather_nd(labels, indices)
-        anchor_state   = keras_retinanet.backend.gather_nd(anchor_state, indices)
+        # compute the divisor: for each image in the batch, we want the number of positive anchors
+
+        # override the -1 labels, since we treat -1/0 the same way for determining the divisor
+        divisor   = keras_retinanet.backend.where(y_true < 0, keras.backend.zeros_like(y_true), y_true)
```
hgaiser: Hmm I always used `keras_retinanet.backend.where(keras.backend.less_equal(y_true, 0))`. Is there a difference?

keras_retinanet/losses.py
```diff
@@ -20,48 +20,62 @@
...
 
         cls_loss = focal_weight * keras.backend.binary_crossentropy(labels, classification)
-        cls_loss = keras.backend.sum(cls_loss)
 
-        # "The total focal loss of an image is computed as the sum
-        # of the focal loss over all ~100k anchors, normalized by the
-        # number of anchors assigned to a ground-truth box."
-        cls_loss = cls_loss / (keras.backend.maximum(1.0, keras.backend.sum(anchor_state)))
-        return cls_loss
+        cls_loss = cls_loss/divisor
```
hgaiser: PEP8

keras_retinanet/losses.py
```diff
@@ -20,48 +20,62 @@
...
-        labels         = keras_retinanet.backend.gather_nd(labels, indices)
-        anchor_state   = keras_retinanet.backend.gather_nd(anchor_state, indices)
+        # compute the divisor: for each image in the batch, we want the number of positive anchors
+
+        # override the -1 labels, since we treat -1/0 the same way for determining the divisor
+        divisor   = keras_retinanet.backend.where(y_true < 0, keras.backend.zeros_like(y_true), y_true)
+        divisor   = keras.backend.max(divisor, axis=2)
+        divisor   = keras.backend.cast(divisor, keras.backend.floatx())
+        # compute the number of positive anchors
+        divisor   = keras.backend.sum(divisor, 1)
```
hgaiser: I prefer the `axis=1` notation for clarity.

keras_retinanet/losses.py
```diff
@@ -20,48 +20,62 @@
...
+        #  ensure we do not divide by 0
+        divisor   = keras.backend.maximum(1.0, divisor)
+        # pad the tensor to have shape (batch_size, 1, 1) for future division
+        divisor   = keras.backend.expand_dims(divisor, axis=1)
```
hgaiser: If you add `keepdims=True` to the `max` and `sum` functions, you don't need these 2 lines.

keras_retinanet/losses.py
```diff
@@ -20,48 +20,62 @@
...
+        cls_loss = cls_loss/divisor
 
-    return _focal
+        # filter out "ignore" anchors
+        anchor_state   = keras.backend.max(labels, axis=2)  # -1 for ignore, 0 for background, 1 for object
+        indices        = keras_retinanet.backend.where(keras.backend.not_equal(anchor_state, -1))
+
+        cls_loss = keras_retinanet.backend.gather_nd(cls_loss, indices)
 
+        return keras.backend.sum(cls_loss)/keras.backend.cast(keras.backend.shape(y_true)[0], keras.backend.floatx())
```
hgaiser: PEP8

keras_retinanet/losses.py
```diff
@@ -20,48 +20,62 @@
...
+        regression        = y_pred
+        regression_target = y_true[:, :, :4]
+        anchor_state      = y_true[:, :, 4]
+
+        # compute the divisor: for each image in the batch, we want the number of positive anchors
+        divisor = keras_retinanet.backend.where(keras.backend.equal(anchor_state, 1), anchor_state, keras.backend.zeros_like(anchor_state))
+        divisor   = keras.backend.sum(divisor, axis=1)
+        divisor   = keras.backend.maximum(1.0, divisor)
+
+        # pad the tensor to have shape (batch_size, 1, 1) for future division
```
hgaiser: Same here regarding the `keepdims`.

keras_retinanet/losses.py
```diff
@@ -73,10 +87,15 @@ def _smooth_l1(y_true, y_pred):
             0.5 * sigma_squared * keras.backend.pow(regression_diff, 2),
             regression_diff - 0.5 / sigma_squared
         )
-        regression_loss = keras.backend.sum(regression_loss)
+        regression_loss = regression_loss /divisor
```
hgaiser: PEP8

keras_retinanet/losses.py
hgaiser: Also, isn't this basically `keras.backend.mean` ?

keras_retinanet/losses.py
```diff
@@ -73,10 +87,15 @@ def _smooth_l1(y_true, y_pred):
...
+
+        # filter out "ignore" anchors
+        indices         = keras_retinanet.backend.where(keras.backend.equal(anchor_state, 1))
+        regression_loss = keras_retinanet.backend.gather_nd(regression_loss, indices)
+
+        # divide loss by the batch size
+        regression_loss = keras.backend.sum(regression_loss) / keras.backend.cast(keras.backend.shape(y_true)[0], keras.backend.floatx())
```
hgaiser: `keras.backend.mean` ?

keras_retinanet/losses.py
hgaiser: So this divides the loss of shape `(batch_size, num_anchors, 1)` by the number of assigned boxes, shaped `(batch_size, 1, 1)` ? Didn't now such operations were possible.

keras_retinanet/losses.py
yhenon: this is definitely a bug!

keras_retinanet/losses.py
yhenon: Indeed, here's a reference: https://www.tensorflow.org/performance/xla/broadcasting

keras_retinanet/losses.py
yhenon: No, if you follow the code closely enough you'll see that `cls_loss` doesn't have `batch_size` entries, it has number of entries equal to the number of positive anchors.

yhenon:
This commit should address the issues:
- replace `<` with `lesser_equal` as it is incorrect
- pep8 compliance
- reduce occurence of `expand_dims` by using `keepdims`
- note that dividing by the batch_size remains as is, as it not equivalent to taking `keras.backend.mean`

hgaiser:
LGTM, thanks!

PR #80, opened 2017-11-20
Split next() in several methods.

hgaiser:
This PR splits the `next()` method from `Generator` into several methods so that they can be overridden in subclasses. This is necessary to generate targets for "other" (non-regression, non-labels) data.

### DIFF
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -22,7 +22,7 @@ import time
 import keras
 
 from keras_retinanet.utils.image import preprocess_image, resize_image, random_transform
-from keras_retinanet.utils.anchors import anchor_targets
+from keras_retinanet.utils.anchors import anchor_targets_bbox
 
 
 class Generator(object):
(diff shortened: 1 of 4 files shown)


keras_retinanet/preprocessing/generator.py
```diff
@@ -149,11 +131,29 @@ def next(self):
...
+        self,
+        image_shape,
+        boxes,
+        num_classes,
+        mask_shape=None,
+        negative_overlap=0.4,
+        positive_overlap=0.5,
+        **kwargs
+    ):
+        return anchor_targets(image_shape, boxes, num_classes, mask_shape, negative_overlap, positive_overlap)
```
de-vri-es: Might rename the free function `anchor_targets` to reflect that it is used for bounding box regression (with labels).

keras_retinanet/preprocessing/generator.py
```diff
@@ -167,4 +167,34 @@ def next(self):
             labels_batch[index, ...]     = labels
             regression_batch[index, ...] = regression
 
-        return image_batch, [regression_batch, labels_batch]
+        return [regression_batch, labels_batch]
+
+    def compute_batches(self, group_index):
```
de-vri-es: The name is slightly misleading. As it is it sounds like it computes the division into batches, but it computes the input and output for a batch.

Some possible alternatives:
 * `compute_batch_input_output`
 * `compute_input_output_batch`
 * `compute_input_output`
 * `compute_batch_blobs`

keras_retinanet/preprocessing/generator.py
hgaiser: Done (renamed to `anchor_targets_bbox`).

keras_retinanet/preprocessing/generator.py
hgaiser: Renamed to `compute_input_output`.

PR #89, opened 2017-11-21
Regression normalization.

hgaiser:
This PR normalizes the values of the bounding box regression (ie subtract mean and divide by standard deviation). Currently, logical mean and std values are given by default, but maybe in the future it could be interesting to generate these based on the data from the `Generator`.


keras_retinanet/losses.py
```diff
@@ -64,16 +63,15 @@ def smooth_l1(sigma=3.0):
...
         regression        = y_pred
         regression_target = y_true[:, :, :4]
         anchor_state      = y_true[:, :, 4]
 
-        # compute the divisor: for each image in the batch, we want the number of positive anchors
-        divisor = keras_retinanet.backend.where(keras.backend.equal(anchor_state, 1), anchor_state, keras.backend.zeros_like(anchor_state))
-        divisor   = keras.backend.sum(divisor, axis=1, keepdims=True)
-        divisor   = keras.backend.maximum(1.0, divisor)
+        # compute the divisor: for each image in the batch, we want the number of positive and negative anchors
+        divisor = keras_retinanet.backend.where(keras.backend.not_equal(anchor_state, -1), keras.backend.ones_like(anchor_state), keras.backend.zeros_like(anchor_state))
```
yhenon: I have doubts regarding this change. This will have the effect of dividing the regression loss by (n_pos_anchors + n_neg_anchors). But only the positive anchors contribute to the loss. This effectively means we are no longer taking the mean regression loss.

keras_retinanet/backend/common.py
```diff
@@ -20,7 +20,12 @@
 import numpy as np
 
 
-def bbox_transform_inv(boxes, deltas):
+def bbox_transform_inv(boxes, deltas, mean=None, std=None):
+    if mean is None:
+        mean = [0, 0, 0, 0]
+    if std is None:
+        std = [0.1, 0.1, 0.2, 0.2]
```
yhenon: I think this approach is good for now. Computing the statistics on the dataset adds some unnecessary complexity.

keras_retinanet/utils/anchors.py
```diff
@@ -174,6 +180,9 @@ def bbox_transform(anchors, gt_boxes):
     targets = np.stack((targets_dx, targets_dy, targets_dw, targets_dh))
     targets = targets.T
 
+    for i in range(4):
+        targets[:, i] = (targets[:, i] - mean[i]) / std[i]
+
```
yhenon: I don't think this loop is needed here, numpy broadcasting rules cover this situation. This works just fine:
```
targets = (targets - mean) / std
```

keras_retinanet/losses.py
hgaiser: I see your point. The change was made because it seems to be what is happening in `py-faster-rcnn` too. Do you think it's better to diverge from `py-faster-rcnn` here? It does mean the regression loss is relative high (around 0.9 when converged). Other than that it shouldn't really matter.

keras_retinanet/utils/anchors.py
hgaiser: I'll try it out, thanks!

hgaiser:
@yhenon I updated the PR. What do you think we should do about the rescaling of the loss by `num_pos` or by `num_pos` + `num_neg` ?

PR #95, opened 2017-11-22
Fix race condition in preprocessing.generator.Generator

de-vri-es:
`Generator` correctly used `self.group_index` only while holding a lock. However, the index was kept after releasing the lock, and the list that the index points into could be reshuffled by another thread.

This PR changes `Generator` to store a reference to the group instead of an index into the group list.


hgaiser:
LGTM, thanks!

PR #96, opened 2017-11-22
Improve error messages of CSVGenerator

de-vri-es:
This PR aims to improve the error messages when CSV files contain invalid data.

Don't merge yet, I want to add unit tests first to make sure I didn't break it. It can already be reviewed though.

Maybe we can prevents merge conflicts when someone adds the option to load images without annotations. Shouldn't be difficult to resolve them if they do occur though.

### DIFF
--- a/keras_retinanet/preprocessing/csv_generator.py
+++ b/keras_retinanet/preprocessing/csv_generator.py
@@ -23,6 +23,7 @@ from PIL import Image
 from six import raise_from
 
 import csv
+import sys
 
 
 def _parse(value, function, fmt):
@@ -66,6 +67,12 @@ def _read_annotations(csv_reader, classes):
         x2 = _parse(x2, int, 'line {}: malformed x2: {{}}'.format(line))
         y2 = _parse(y2, int, 'line {}: malformed y2: {{}}'.format(line))
 
+        # Check that the bounding box is valid.
+        if x2 <= x1:
+            raise ValueError('line {}: x2 ({}) must be higher than x1 ({})'.format(line, x2, x1))
+        if y2 <= y1:
+            raise ValueError('line {}: y2 ({}) must be higher than y1 ({})'.format(line, y2, y1))
+
         # check if the current class name is correctly present
         if class_name not in classes:
             raise ValueError('line {}: unknown class name: {}'.format(line, class_name))
@@ -81,10 +88,10 @@ def _open_for_csv(path):
     """
     Open a file with flags suitable for csv.reader.
 
-    This is different for python2 this means with mode 'rb',
+    This is different for python2 it means with mode 'rb',
     for python3 this means 'r' with "universal newlines".
     """
-    if bytes == str:
+    if sys.version_info[0] < 3:
         return open(path, 'rb')
     else:
         return open(path, 'r', newline='')
--- a/keras_retinanet/preprocessing/csv_generator.py
+++ b/keras_retinanet/preprocessing/csv_generator.py
@@ -15,13 +15,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from keras_retinanet.preprocessing.generator import Generator
+from .generator import Generator
+from ..utils.image import read_image_bgr
 
 import numpy as np
 from PIL import Image
 from six import raise_from
 
-import cv2
 import csv
 
 
@@ -142,7 +142,7 @@ class CSVGenerator(Generator):
 
     def load_image(self, image_index):
         path = self.image_names[image_index]
-        return cv2.imread(path)
+        return read_image_bgr(path)
 
     def load_annotations(self, image_index):
         path   = self.image_names[image_index]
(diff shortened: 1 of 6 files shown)


keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -31,40 +78,28 @@ def __init__(
...
-        self.csv_data_file        = csv_data_file
-
-        self.image_names          = []
-        self.image_data           = {}
+        self.image_names = []
+        self.image_data  = {}
 
         # parse the provided class file
-        self.classes = {}
-        with open(csv_class_file, 'r') as f_class_in:
```
yhenon: Shoud be `'rb'` for windows users.

de-vri-es: Updated, thanks.

hgaiser: But 'r' for Unix ?

de-vri-es: On unix systems 'b' doesn't do anything.

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -122,7 +121,8 @@ def image_aspect_ratio(self, image_index):
 
     def load_image(self, image_index):
         path = self.image_names[image_index]
-        return cv2.imread(path)
+        # Return image with channels reversed (BGR instead of RGB).
+        return np.asarray(Image.open(path))[:, :, ::-1]
```
hgaiser: I don't mind if this is done for all generators. I'll make a PR for removing the opencv dependency in other generators and in the data augmentation code.

de-vri-es: I intended to add a utility function `utils.image.read_image_bgr()`, but then I noticed that there is still a opencv dependency there even if reading images is done with PIL.

Still, we should probably put it there, otherwise we're going to have `np.asarray(Image.open(...))[:, :, ::-1]` all over the place.

de-vri-es: Moved it to utils now.

keras_retinanet/preprocessing/csv_generator.py
hgaiser: Seems that the documentation is different for python2/3. In the [documentation](https://docs.python.org/2/library/csv.html#csv.reader) for python2 they mention:
```
If csvfile is a file object, it must be opened with the ‘b’ flag on platforms where that makes a difference.
```
Whereas in the python3 [documentation](https://docs.python.org/3/library/csv.html#csv.reader) they say:
```
If csvfile is a file object, it should be opened with newline=''.
```

In addition, a small test on python3, testing `'r'` vs `'rb'` shows the latter doesn't work, with the following error:

```
Error: iterator should return strings, not bytes (did you open the file in text mode?)
```

de-vri-es: I think this is fixed now. Could you see what you think of it?

hgaiser: Should the python3 version not still be opened with `'r'`? Also maybe add a comment for why this is necessary. Other than that, LGTM (unfortunately though).

de-vri-es: 'r' is the default mode for opening file: https://docs.python.org/3/library/functions.html#open

hgaiser: Maybe add it anyway for clarity? And comments on why that function is there would be helpful too.

de-vri-es: I added a docstring and the mode.

hgaiser: Then it LGTM. @yhenon ?

yhenon: I think the way suggested here is clearer actually:  https://stackoverflow.com/questions/5180555/python-2-and-3-csv-reader

keras_retinanet/utils/image.py
```diff
@@ -19,6 +19,12 @@
 import time
 import numpy as np
 import cv2
+import PIL
+
+
+def read_image_bgr(path):
+    image = np.asarray(PIL.Image.open(path).convert('RGB'))
+    return image[:, :, ::-1]
```
yhenon: We should avoid this (it might be responsible for #97), or at least make a copy afterwards.
This creates a view:
```
>>> image = np.asarray(PIL.Image.open(path).convert('RGB'))
>>> image.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  UPDATEIFCOPY : False
>>> image = np.asarray(PIL.Image.open('crop.png').convert('RGB'))[:,:,::-1]
>>> image.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  UPDATEIFCOPY : False
```

hgaiser: I don't think that issue was caused by this because `master` currently still uses `cv2.imread`, but I get your point. @de-vri-es do you mind making the suggested adjustments? My preference would be to use `copy()`, since `opencv` does that internally anyway.

PR #98, opened 2017-11-23
A simple clarification for COCO images setup

eakbas:
I downloaded the `val2017` images and put them under `/srv/datasets/COCO` which did not work. The problem was `cv2.imread` was returning `None` without any descriptive message. Later I realized that the library was looking for images under `/srv/datasets/COCO/images/val2017` not under `/srv/datasets/COCO/val2017`

I added a simple comment to clarify this.

### DIFF
--- a/examples/ResNet50RetinaNet - COCO 2017.ipynb
+++ b/examples/ResNet50RetinaNet - COCO 2017.ipynb
@@ -88,7 +88,7 @@
     "\n",
     "# create a generator for testing data\n",
     "val_generator = CocoGenerator(\n",
-    "    '/srv/datasets/COCO',\n",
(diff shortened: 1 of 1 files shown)


hgaiser:
I agree, the error messaging needs improvement when the image cannot be found. https://github.com/fizyr/keras-retinanet/pull/102 should address that as it switches from opencv (which doesn't error if an image cannot be found) to PIL (which gives clear errors saying why an image could not be opened). I hope that PR will prevent errors like this from happening. For now I'll close this PR, as the https://github.com/fizyr/keras-retinanet/pull/102 hopefully provides a clearer error (and therefore solution).

PR #102, opened 2017-11-23
Reduce dependency on OpenCV.

hgaiser:
This PR reduces the dependency on OpenCV.

Unfortunately, OpenCV is just really fast (order of 5 times faster than PIL) in resizing an image, so the dependency is kept for `cv2.resize`.


de-vri-es:
Note that opencv has caused some compatibility problems in the past. Also, if `cv2.imread` fails it does so silently (retunring `None`), whereas `PIL.Image.open()` will raise an error describing what went wrong.

So +1 from me.

PR #103, opened 2017-11-23
Check for invalid bounding boxes in the generator.

hgaiser:
This PR adds a check for invalid bounding boxes (where `x2 < x1` or `y2 < y1`).

Note that it prints a warning about these incorrect bounding boxes, filters them from the data and continues running.

### DIFF
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -86,9 +86,7 @@ class Generator(object):
                 (annotations[:, 2] <= annotations[:, 0]) |
                 (annotations[:, 3] <= annotations[:, 1]) |
                 (annotations[:, 0] < 0) |
-                (annotations[:, 1] < 0) |
-                (annotations[:, 2] <= 0) |
-                (annotations[:, 3] <= 0)
+                (annotations[:, 1] < 0)
             )[0]
             if len(invalid_indices):
                 warnings.warn('Image with id {} contains the following invalid boxes: {}.'.format(
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -18,6 +18,7 @@ import numpy as np
 import random
 import threading
 import time
+import warnings
 
 import keras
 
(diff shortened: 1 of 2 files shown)


hgaiser:
> It would be really great to have a unit test for this though.

Done. Can you re-review ?

keras_retinanet/preprocessing/generator.py
```diff
@@ -53,28 +54,50 @@ def __init__(
...
+            assert(isinstance(annotations, np.ndarray)), '\'load_annotations\' should return a list of numpy arrays, received: {}'.format(type(annotations))
+
+            # test x2 < x1 | y2 < y1 | x1 < 0 | y1 < 0 | x2 <= 0 | y2 <= 0
+            invalid_indices = np.where(
+                (annotations[:, 2] <= annotations[:, 0]) |
+                (annotations[:, 3] <= annotations[:, 1]) |
+                (annotations[:, 0] < 0) |
+                (annotations[:, 1] < 0) |
+                (annotations[:, 2] <= 0) |
+                (annotations[:, 3] <= 0)
```
de-vri-es: Considering that you check `x2 <= x1` and `x1 < 0`, checking `x2 <= 0` seems a bit reduntant.

keras_retinanet/preprocessing/generator.py
```diff
@@ -53,28 +54,50 @@ def __init__(
...
+            )[0]
+            if len(invalid_indices):
+                warnings.warn('Image with id {} contains the following invalid boxes: {}.'.format(
+                    group[index],
+                    [annotations[invalid_index, :] for invalid_index in invalid_indices]
+                ))
```
de-vri-es: +1 for listing the indices in the warning

tests/preprocessing/test_generator.py
```diff
@@ -0,0 +1,133 @@
...
+                [  0,   0,  10,  10],
+                [150, 150, 350, 350]
+            ]),
+        ]
+        expected_annotations_group = input_annotations_group
+
+        simple_generator = SimpleGenerator(input_annotations_group)
+        annotations_group = simple_generator.load_annotations_group(simple_generator.groups[0])
+
+        np.testing.assert_equal(expected_annotations_group, annotations_group)
```
de-vri-es: Normally I prefer to perform the calculation in the assertion itself, since most testing libraries can then print the calculation that failed. I don't know if `np.testing.assert_equal()` can do that though. Probably not, since it is just a function call.

tests/preprocessing/test_generator.py
```diff
@@ -0,0 +1,133 @@
...
+                [10, 10, 100, 100]
+            ])
+        ]
+
+        simple_generator = SimpleGenerator(input_annotations_group)
+        # expect a UserWarning
+        with pytest.warns(UserWarning):
+            annotations_group_0 = simple_generator.load_annotations_group(simple_generator.groups[0])
+            annotations_group_1 = simple_generator.load_annotations_group(simple_generator.groups[1])
+            annotations_group_2 = simple_generator.load_annotations_group(simple_generator.groups[2])
```
de-vri-es: Each of these statements should generate a warning right? The way it's written now I think the test only checks that one of them generates a warning.

hgaiser: Good point, that's true.

tests/preprocessing/test_generator.py
hgaiser: numpy shows that there is a x% mismatch between expected and result, and prints the arrays.

keras_retinanet/preprocessing/generator.py
hgaiser: True that.

hgaiser:
I'll wait with merging this until #96 gets merged, otherwise Travis starts complaining..

tests/preprocessing/test_generator.py
de-vri-es: What I mean is that if you do something like `assert sqrt(4) == 3`, pytest will print `sqrt(4) == 3`, which is more informative when a test fails than `assert expected == actual`.

But in this case I don't think numpy can do the same, so there probably isn't a difference.

keras_retinanet/preprocessing/generator.py
```diff
@@ -53,28 +54,48 @@ def __init__(
...
+
+        # test all annotations
+        for index, annotations in enumerate(annotations_group):
+            assert(isinstance(annotations, np.ndarray)), '\'load_annotations\' should return a list of numpy arrays, received: {}'.format(type(annotations))
+
+            # test x2 < x1 | y2 < y1 | x1 < 0 | y1 < 0 | x2 <= 0 | y2 <= 0
+            invalid_indices = np.where(
+                (annotations[:, 2] <= annotations[:, 0]) |
+                (annotations[:, 3] <= annotations[:, 1]) |
+                (annotations[:, 0] < 0) |
```
yhenon: If we check for boxes with x1<0, shouldn't we also check for images with x2>image_width? Of course this is more difficult since we need access to the image metadata

hgaiser: Would be nice, yeah, but I'd like to avoid a call similar to `image_aspect_ratio` so maybe it's better to skip this check.

PR #111, opened 2017-11-27
Multi gpu training

hgaiser:
This PR enables the use of multiple GPUs to train RetinaNet.

Note, for the callbacks, PR https://github.com/fchollet/keras/pull/8614 of Keras is required. We should probably hold off merging this until that or something similar gets merged into Keras.


de-vri-es:
There's a reasonably high amount of code duplication in the scripts. Could we put something general in `utils` to reuse?

hgaiser:
I'm actually currently looking into merging them all in one script. Basically, the only thing that is different is the data generator..

de-vri-es:
That also sounds nice.

hgaiser:
@yhenon @de-vri-es could you guys review this PR again? Currently it evolved into the following:

- Change `Anchors` and `RegressBoxes` layers to allow `batch_size > 1` in these layers.
- Move all training scripts into one script.
- Move `train.py` and `evaluate_coco.py` to a separate folder called `scripts`.
- Add the option to use multi-gpu training in `train.py`. Note that it is tested, but not extensively due to limited resources.

scripts/train.py
```diff
@@ -0,0 +1,206 @@
...
+def create_generators(args):
+    # create image data generator objects
+    train_image_data_generator = keras.preprocessing.image.ImageDataGenerator(
+        horizontal_flip=True,
+    )
+    val_image_data_generator = keras.preprocessing.image.ImageDataGenerator()
+
+    if args.dataset_type == 'coco':
+        # import here to prevent unnecessary dependency on cocoapi
+        from keras_retinanet.preprocessing.coco import CocoGenerator
```
de-vri-es: +1

scripts/train.py
```diff
@@ -0,0 +1,206 @@
...
+    coco_parser = subparsers.add_parser('coco')
+    coco_parser.add_argument('coco_path', help='Path to dataset directory (ie. /tmp/COCO).')
+
+    pascal_parser = subparsers.add_parser('pascal')
+    pascal_parser.add_argument('pascal_path', help='Path to dataset directory (ie. /tmp/VOCdevkit).')
+
+    csv_parser = subparsers.add_parser('csv')
+    csv_parser.add_argument('csv_path', help='Path to CSV file for training.')
+    csv_parser.add_argument('classes_path', help='Path to a CSV file containing class label mapping.')
+    csv_parser.add_argument('--val_path', help='Path to CSV file for validation (optional).')
```
de-vri-es: Should be `--val-path`?

Also, `csv_path` isn't very descriptive since `classes_path` also designates a CSV file. How abbout `annotations` and `classes`? That together with a good description should be clear enough.

tests/backend/test_common.py
```diff
@@ -42,10 +42,10 @@ def test_bbox_transform_inv():
         [0   , 0  , 0   , -0.3],
         [0   , 0  , 0.2 , 0.2 ],
         [0.1 , 0.2, -0.3, 0.4 ],
-    ])
+    ]])
```
de-vri-es: Considering that the expected output for the tests changed, does this mean the output format is incompatible with versions before this PR?

scripts/train.py
hgaiser: :+1:

tests/backend/test_common.py
hgaiser: Only if someone directly uses `keras_retinanet.backed.bbox_transform_inv`, other than that backwards compatibility is maintained. Before this PR the `RegressBoxes` layer would select the first index in the batch dimension, pass that blob to `bbox_transform_inv`, extend the dimensions by one to get a "batch dimension" again and output that blob. This PR changes the `RegressBoxes` such that `batch_size > 1` is handled correctly, but this also means `bbox_transform_inv` should handle that.

de-vri-es: 👍

hgaiser:
Made the requested changes, all is good now?

de-vri-es:
All is good.

PR #116, opened 2017-11-29
modify_group_shuffle

fangwudi:
modify shuffle_groups to make it more clear: use it in start of epoch and delete shuffle in initiation


hgaiser:
Thanks!

de-vri-es:
I meant "good" of course, not "goo" =]

Issue #117, opened 2017-11-29
Relative image paths in CSVGenerator


de-vri-es:
Relative image paths in the CSV generator are resolved with respect to the current working directory.

However, in the interest of having relocatable datasets, it might make sense to resolve them with respect to the directory of the CSV file.

@hgaiser , @yhenon , what do you guys think?

hgaiser:
The COCO and Pascal generators work slightly different. They expect a working directory and the name of a file listing which images to use. Of course there you can expect the (for COCO for instance) the annotations to be in the `annotations` folder since that's how the COCO dataset looks like. But perhaps the CSV generator could do something similar by looking for the annotations / images / class file relative to the given working directory?

de-vri-es:
I would prefer not to hardcode the names of the csv files, if that is what you propose. That would reduce the flexibility of the generator. The pascal and VOC generator aren't flexible at all, but the CSV generator is.

Without reducing flexibility we could have three parameters:
* annotations file
* class file
* base directory

Then if directory isn't given it could be deduced from the annotations file, and if directory is given we can deduce a default for annotations file and class file. That does get a bit more complicated to explain though.

Simply having relative paths in a file be relative to the directory of that file is quite a bit easier to explain, although less flexible.

de-vri-es:
Initial implementation at the [csv-relative-paths](https://github.com/fizyr/keras-retinanet/tree/csv-relative-paths) branch. It should be relatively straight-forward to expose the `base_dir` argument to users.

de-vri-es:
Indeed :)

de-vri-es:
Closing this in favour of #125

PR #125, opened 2017-11-30
Interpret relative image paths as relative to CSV file in CSVGenerator.

de-vri-es:
Lets continue #117 here.

To sum up, this PR would make all relative image paths in a CSV file be interpreted relative to some `base_dir`. It can be specified as a constructor argument to `CSVGenerator`, but it defaults to the directory containing the annotations file.

This does *not* interpret the paths to the annotations file or class file relative to the `base_dir` if it is given. I would always expect paths I specify on the command line to be relative to the current working directory, and I think it would be bad to break that expectation.

### DIFF
--- a/keras_retinanet/preprocessing/csv_generator.py
+++ b/keras_retinanet/preprocessing/csv_generator.py
@@ -24,6 +24,7 @@ from six import raise_from
 
 import csv
 import sys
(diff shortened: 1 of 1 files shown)


hgaiser:
LGTM!

PR #131, opened 2017-11-30
Support images without annotations in the CSV generator.

de-vri-es:
This PR allows images without annotations in the CSV generator. They can be listed with `x1`, `y1`, `x2`, `y2` and `class_name` empty:
```
path/to/image.jpg,,,,,
```

I also considered listing just the image and no trailing commas, but at least LibreOffice adds them when saving to CSV.

Unit test and README updated. I moved the format of the CSV files to their own section because it was getting a bit long.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,34 +22,13 @@ For training on [MS COCO](http://cocodataset.org/#home), run:
 python examples/train_coco.py <path to MS COCO>
 ```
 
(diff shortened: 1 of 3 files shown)


de-vri-es:
Since there are no objections, I'm merging. Thanks for the quick feedback =)

PR #132, opened 2017-11-30
Use relative imports for all package code.

de-vri-es:
Using absolute imports for other modules within the same package is bad practise. It can break in strange ways when installed packages are mixed with local versions.

This PR replaces *all* absolute imports for modules in the package with relative imports. Scripts sadly don't support relative imports, so the examples can not be fixed this way.

### DIFF
--- a/keras_retinanet/backend/__init__.py
+++ b/keras_retinanet/backend/__init__.py
@@ -1,27 +1,4 @@
 import os
 
+from .dynamic import *
 from .common import *
-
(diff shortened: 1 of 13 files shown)


fangwudi:
I like it

de-vri-es:
I tested training on COCO with `steps_per_epoch=10`, and it got past the first epoch with both python2 and python3. So it seems to still work correctly.

hgaiser:
LGTM!

PR #133, opened 2017-12-01
fix args is not defined

wassname:


hgaiser:
Thanks for the PR, but I prefer to fix it the other way around:
https://github.com/fizyr/keras-retinanet/commit/81eb47472e66c200541805a6eaf875890137f390

wassname:
No probs

PR #140, opened 2017-12-01
Pascal VOC generator image names reader fix

mxvs:
The current version of the PASCAL VOC generator does not process the format correctly, more specifically the image set files containing the image names are not handled properly.

Example lines from one of the image set files files:
```
2008_000003 -1
2008_000002 -1
2008_000003 -1
2008_000007 -1
2008_000008 -1
```

Only the first part is the actual image name that needs to be processed, running the current generator against the official PASCAL VOC dataset generates the following error(s):

```
FileNotFoundError: [Errno 2] No such file or directory: '/var/dl-data/VOCdevkit/VOC2012/JPEGImages/2008_000002 -1.jpg'
```

As you can see the second part of the line, the `-1` is also being read and considered as part of the image name.

This small patch fixes this issue by splitting the string and ensuring only the actual filenames are being parsed.

### DIFF
--- a/keras_retinanet/preprocessing/pascal_voc.py
+++ b/keras_retinanet/preprocessing/pascal_voc.py
@@ -81,7 +81,7 @@ class PascalVocGenerator(Generator):
         self.data_dir             = data_dir
         self.set_name             = set_name
         self.classes              = classes
-        self.image_names          = [l.strip() for l in open(os.path.join(data_dir, 'ImageSets', 'Main', set_name + '.txt')).readlines()]
+        self.image_names          = [l.strip().split(None, 1)[0] for l in open(os.path.join(data_dir, 'ImageSets', 'Main', set_name + '.txt')).readlines()]
         self.image_extension      = image_extension
         self.skip_truncated       = skip_truncated
         self.skip_difficult       = skip_difficult
--- a/keras_retinanet/preprocessing/pascal_voc.py
+++ b/keras_retinanet/preprocessing/pascal_voc.py
@@ -81,7 +81,7 @@ class PascalVocGenerator(Generator):
         self.data_dir             = data_dir
         self.set_name             = set_name
         self.classes              = classes
-        self.image_names          = [l.strip().split(None, 1)[0] for l in open(os.path.join(data_dir, 'ImageSets', 'Main', set_name + '.txt')).readlines()]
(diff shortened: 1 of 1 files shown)


yhenon:
Thanks for the PR. Looking over the pascal voc dataset, it seems to be like this format:
```
2008_000003 -1
2008_000002 -1
2008_000003 -1
2008_000007 -1
2008_000008 -1
```
is only used for the class specific image sets, but not for the overall image sets. Is the intent of this PR the ability to train on just one of the classes by specifying for example `chair_train.txt` as your image set?

keras_retinanet/preprocessing/pascal_voc.py
```diff
@@ -81,7 +81,7 @@ def __init__(
         self.data_dir             = data_dir
         self.set_name             = set_name
         self.classes              = classes
-        self.image_names          = [l.strip() for l in open(os.path.join(data_dir, 'ImageSets', 'Main', set_name + '.txt')).readlines()]
+        self.image_names          = [l.strip().split(None, 1)[0] for l in open(os.path.join(data_dir, 'ImageSets', 'Main', set_name + '.txt')).readlines()]
```
de-vri-es: I would prefer to write this as `str.split(maxsplit=1)` instead of `str.split(None, 1)`. I think the first version is much easier to understand without referring to the `str` documentation.

Also, what does the -1 even mean here? Why is it there?

hgaiser:
Any update on this ?

hgaiser:
I'll close this PR due to inactivity. It needs more feedback. Feel free to re-open if necessary.

mxvs:
Hello,

Sorry I was out of town and unable to reply. Its correct as mentioned by @yhenon that the original dataset containing ALL images does not have the additional character in the textfile. However when you want to train on a specific class subset it doesn't work with the current iterator.

More importantly when creating your OWN dataset using tools such as https://github.com/Microsoft/VoTT you also get the additional 1 or -1 and the generator does not currently work on files generated by this or other tools.

My change also does not break the existing code at all, it simply allows for the parsing of additional image sets in the original dataset or use datasets generated by tools like the one from Microsoft.

I also updated the change to use str.split(maxsplit=1) as indicated by @de-vri-es

de-vri-es:
I still wonder what this number means. Also, with this change files will not be able to have spaces in their names anymore. However, that seems to be a limit of the format itself, so there is little we can do about it here.

de-vri-es:
Thanks for the PR!

PR #148, opened 2017-12-03
Extend tests on annotations in Generator.

hgaiser:
This PR extends the tests on annotations by also checking if they don't exceed the image shape.

### DIFF
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -75,27 +75,33 @@ class Generator(object):
         raise NotImplementedError('load_annotations method not implemented')
 
     def load_annotations_group(self, group):
-        annotations_group = [self.load_annotations(image_index) for image_index in group]
+        return [self.load_annotations(image_index) for image_index in group]
 
+    def filter_annotations(self, image_group, annotations_group, group):
         # test all annotations
-        for index, annotations in enumerate(annotations_group):
+        for index, (image, annotations) in enumerate(zip(image_group, annotations_group)):
             assert(isinstance(annotations, np.ndarray)), '\'load_annotations\' should return a list of numpy arrays, received: {}'.format(type(annotations))
 
-            # test x2 < x1 | y2 < y1 | x1 < 0 | y1 < 0 | x2 <= 0 | y2 <= 0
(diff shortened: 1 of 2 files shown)


keras_retinanet/preprocessing/generator.py
```diff
@@ -75,27 +75,33 @@ def load_annotations(self, image_index):
         raise NotImplementedError('load_annotations method not implemented')
 
     def load_annotations_group(self, group):
-        annotations_group = [self.load_annotations(image_index) for image_index in group]
+        return [self.load_annotations(image_index) for image_index in group]
 
+    def check_images_annotations(self, image_group, annotations_group, group):
```
de-vri-es: `check` implies that the function returns true/false. However, it filters the annotations. How about `filter_annotations` or similar?

tests/preprocessing/test_generator.py
```diff
@@ -76,16 +76,19 @@ def test_simple_filter(self):
...
                 [  0,   0, 10, 10],
             ]),
         ]
 
         simple_generator = SimpleGenerator(input_annotations_group)
+        annotations_group = simple_generator.load_annotations_group(simple_generator.groups[0])
         # expect a UserWarning
         with pytest.warns(UserWarning):
-            annotations_group = simple_generator.load_annotations_group(simple_generator.groups[0])
+            image_group, annotations_group = simple_generator.check_images_annotations([input_image], annotations_group, simple_generator.groups[0])
```
de-vri-es: Should we consider `check_image_annotations` to be a public interface? Or is it a private/protected implementation detail and/or extension point for subclasses?

In the latter case we should probably not test it directly but test the public interface instead. Or at the very least test that the public interface also filters properly.

keras_retinanet/preprocessing/generator.py
hgaiser: Makes sense, I agree.

tests/preprocessing/test_generator.py
hgaiser: How do you propose to do that though? The public interface is through the `next()` function, but that returns anchor targets. Should we test on those anchor targets?

de-vri-es: As discussed in person: What we could do is have a dataset with two annotations with different labels, where one of the annotations has to be discarded by the filter. Then let the generator generate two values and assert that both of them have the label of the valid annotation.

This tests at the very least that the filtering is done.

hgaiser:
Made the requested changes, can you check?

PR #149, opened 2017-12-04
random_transform should output same box size

wassname:
When I test it random transform clip the boxes, here is a notebook showing/plotting the problem and solution https://gist.github.com/wassname/36e9cb71286e37f724c0d0169988eca7


hgaiser:
Nice find, thanks!

hgaiser:
I solved it a bit differently in https://github.com/fizyr/keras-retinanet/commit/7ec912558ce328a8b0242e49ba3025965d5665bf . The boxes should be defined as an open interval (so for example `[x1, x2)` ).

With this PR, if a box is defined with `x2 = image.shape[1]` for example, the augmented box would still become `x2 = image.shape[1] - 1`.

With https://github.com/fizyr/keras-retinanet/commit/7ec912558ce328a8b0242e49ba3025965d5665bf , it would become `x2 = image.shape[1]`.

PR #158, opened 2017-12-05
testing code for csv

HuangBo-Terraloupe:
I add a ipython notebook for testing model training for csv.


hgaiser:
Thank you for your PR, but I'm a bit hesitant to accept this since it's an extra example that would need to be maintained. Plus notebooks are a bit more annoying to maintain than regular python code. I'm hoping at one point we can do something like what is done for the training scripts in https://github.com/fizyr/keras-retinanet/pull/111, that is merge all examples in a more modular way.

@de-vri-es what do you think?

de-vri-es:
I'm also not a big fan of jupyter notebooks in the repository. It's basically a binary blob with some json around it. Keeping it in sync with changes to the rest of the code means opening it manually. Even simple syntax checks can't be done automatically easily.

A general evaluation script is something I can see value in, but then it should be a regular python script with support for all the generators that we have, like the new training script.

hgaiser:
Alright, then I'm closing this PR.

PR #165, opened 2017-12-06
Create base model for multi_gpu_model on cpu

awilliamson:
Keras documentation for multi_gpu_model recommends creating the template
base model on the CPU to ease weight sharing, and to prevent some OOM
errors. This template will then be shared to multiple GPUs.
It is my understanding that this approach is "better"; whether this results in real-world tangible benefits is up for debate. But this should be considered best practice according to Keras.


hgaiser:
Thank you, looks good to me!

PR #170, opened 2017-12-07
Test training procedure

hgaiser:
This PR adds a unit test for a simple training procedure on a small COCO training dataset.

The intention is to add more tests. For instance also for CSV / Pascal training. Before doing that I'd like to review these changes to see if this is really the "best practice".


keras_retinanet/bin/train.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
```
de-vri-es: Since this file moved into the `keras-retinanet` package we should use relative imports. However, this file can be used as library and as script. That means special care is required to make relative imports work in both cases.

keras_retinanet/bin/train.py
```diff
@@ -188,14 +191,19 @@ def parse_args():
...
-    parser.add_argument('--batch-size', help='Size of the batches.', default=1, type=int)
-    parser.add_argument('--gpu', help='Id of the GPU to use (as reported by nvidia-smi).')
-    parser.add_argument('--multi-gpu', help='Number of GPUs to use for parallel processing.', type=int, default=0)
+    parser.add_argument('--weights',       help='Weights to use for initialization (defaults to ImageNet).', default='imagenet')
+    parser.add_argument('--batch-size',    help='Size of the batches.', default=1, type=int)
+    parser.add_argument('--gpu',           help='Id of the GPU to use (as reported by nvidia-smi).')
+    parser.add_argument('--multi-gpu',     help='Number of GPUs to use for parallel processing.', type=int, default=0)
+    parser.add_argument('--epochs',        help='Number of epochs to train.', type=int, default=50)
+    parser.add_argument('--steps',         help='Number of steps per epoch.', type=int, default=10000)
+    parser.add_argument('--no-evaluation', help='Disable per epoch evaluation.', dest='evaluation', action='store_false')
```
de-vri-es: `--no-evaluation` is kinda double (although inverted) with `--val-annotations` for the `CSVGenerator`. We could instead add `--use-validation` to the COCO subparser to avoid having double options.

tests/test_train.py
```diff
@@ -0,0 +1,38 @@
...
+        '',
+        '--epochs=1',
+        '--steps=1',
+        '--no-snapshots',
+        'coco',
+        'tests/test-data/coco',
+    ]
+
+    # run training / evaluation
+    keras_retinanet.bin.train.main()
```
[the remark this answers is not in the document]
hgaiser: I just tried it, the tests don't have access to globally defined variables.

keras_retinanet/bin/train.py
hgaiser: Would you be so kind as to look into this issue?

keras_retinanet/bin/train.py
hgaiser: Hmm I get it, but I would prefer that for COCO at least the evaluation is on by default. How would you suggest to do this?

awilliamson: You can set mutually exclusive command line args by creating a group under parser.
Eg ```python
parser = argparse.ArgumentParser(prog='PROG')
group = parser.add_mutually_exclusive_group()
group.add_argument('--foo', action='store_true')
group.add_argument('--bar', action='store_false')```

keras_retinanet/bin/train.py
awilliamson: Additionally, what about those using virtualenvs? Would this overwrite the default python binary to execute? If hashbangs are going to be used, should they be consistent with all python files?

de-vri-es: using `/usr/bin/env` to execute python ensures that it is looked up using $PATH. To my knowledge, virtualenvs use $PATH to change the invoked python interpreter, so this shebang should work fine with virtualenv.

keras_retinanet/bin/train.py
de-vri-es: we could simply add `--no-validation` to the Coco subparser instead of the main parser I think.

tests/test_train.py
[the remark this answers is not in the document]
de-vri-es: good to hear :)

keras_retinanet/bin/train.py
de-vri-es: I added two commits to use relative imports in the scripts/modules in bin. I pretty much did what [PEP-366](https://www.python.org/dev/peps/pep-0366/) suggests. Sadly this includes fiddling with `sys.path`.

Relevant excerpt from PEP-366:

>When the main module is specified by its filename, then the `__package__` attribute will be set to None. To allow relative imports when the module is executed directly, boilerplate similar to the following would be needed before the first relative import statement:

```
if __name__ == "__main__" and __package__ is None:
    __package__ = "expected.package.name"
```

> Note that this boilerplate is sufficient only if the top level package is already accessible via `sys.path`. Additional code that manipulates `sys.path` would be needed in order for direct execution to work without the top level package already being importable.

keras_retinanet/bin/train.py
hgaiser: Named it `--no-evaluation`, but yeah.

When all datasets get evaluation code we can change this, for now this will do.

hgaiser:
@de-vri-es I updated this PR with Pascal and CSV tests. Can you check it out?

keras_retinanet/bin/train.py
de-vri-es: `--val-annotations` and `--no-evaluation` isn't very consistent. I think we should use the same term for all generators.

If that term is "evaluation", how about changing `--val-annotations` to simply `--evaluate`? The description can make it clear that the argument should be a CSV annotations file.

tests/bin/test_train.py
```diff
@@ -0,0 +1,75 @@
...
+        '',
+        '--epochs=1',
+        '--steps=1',
+        '--no-snapshots',
+        'coco',
+        'tests/test-data/coco',
+    ]
+
+    # run training / evaluation
+    keras_retinanet.bin.train.main()
```
de-vri-es: We could let `main()` take a list of arguments which defaults to `sys.argv`. Then we could call it with the arguments here directly without setting `sys.argv`. What do you think?

hgaiser:
Yeah, sounds better than what we have now :)

de-vri-es:
> Yeah, sounds better than what we have now :)

Pushed a commit for it. The email reply was put in the main comment thread btw, not in the in-line thread.

keras_retinanet/bin/evaluate_coco.py
```diff
@@ -41,20 +41,22 @@ def get_session():
...
-    return parser.parse_args()
+    return parser.parse_args(args)
 
 
-def main():
+def main(args=None):
     # parse arguments
-    args = parse_args()
+    if args is None:
+        args = sys.argv[1:]
```
de-vri-es: Did it this way (default to `None`, check in body) because otherwise the default value of `args` would be determined at parse time. If `args` defaults to `sys.argv[1:]` directly, updating `sys.argv` and then calling `main` would not change the default value (because it was determined at parse time).

hgaiser:
I don't really have the time for it, but could you update the README and then merge this?

de-vri-es:
Readme updated, merging.

PR #176, opened 2017-12-11
Argparsed snapshot path

awilliamson:
Enables multiple executions as parameter, for hyperparameter search and
increase compatibility for scripting experiment runs.
As discussed with @hgaiser.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -71,11 +71,17 @@ def create_models(num_classes, weights='imagenet', multi_gpu=0):
     return model, training_model, prediction_model
 
 
-def create_callbacks(model, training_model, prediction_model, validation_generator, dataset_type):
+def create_callbacks(model, training_model, prediction_model, validation_generator, dataset_type, snapshot_path):
     callbacks = []
 
     # save the prediction model
-    checkpoint = keras.callbacks.ModelCheckpoint(os.path.join('snapshots', 'resnet50_{dataset_type}_{{epoch:02d}}.h5'.format(dataset_type=dataset_type)), verbose=1)
+    checkpoint = keras.callbacks.ModelCheckpoint(
+        os.path.join(
+            snapshot_path,
+            'resnet50_{dataset_type}_{{epoch:02d}}.h5'.format(dataset_type=dataset_type)
+        ),
+        verbose=1
+    )
     checkpoint = RedirectModel(checkpoint, prediction_model)
     callbacks.append(checkpoint)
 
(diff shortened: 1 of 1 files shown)


scripts/train.py
```diff
@@ -192,6 +194,8 @@ def parse_args():
     parser.add_argument('--batch-size', help='Size of the batches.', default=1, type=int)
     parser.add_argument('--gpu', help='Id of the GPU to use (as reported by nvidia-smi).')
     parser.add_argument('--multi-gpu', help='Number of GPUs to use for parallel processing.', type=int, default=0)
+    parser.add_argument('--snapshot-path', help='Path to store snapshots of models during training (defaults to '
```
hgaiser: I'm fine with this being one line.

scripts/train.py
```diff
@@ -218,7 +222,8 @@ def parse_args():
     print(model.summary())
 
     # create the callbacks
-    callbacks = create_callbacks(model, training_model, prediction_model, validation_generator, args.dataset_type)
+    callbacks = create_callbacks(model, training_model, prediction_model, validation_generator, args.dataset_type,
```
hgaiser: Same here.

scripts/train.py
```diff
@@ -71,11 +71,13 @@ def create_models(num_classes, weights='imagenet', multi_gpu=0):
...
 
-def create_callbacks(model, training_model, prediction_model, validation_generator, dataset_type):
+def create_callbacks(model, training_model, prediction_model, validation_generator, dataset_type, snapshot_path):
     callbacks = []
 
     # save the prediction model
-    checkpoint = keras.callbacks.ModelCheckpoint(os.path.join('snapshots', 'resnet50_{dataset_type}_{{epoch:02d}}.h5'.format(dataset_type=dataset_type)), verbose=1)
+    checkpoint = keras.callbacks.ModelCheckpoint(os.path.join(snapshot_path,
+                                                              'resnet50_{dataset_type}_{{epoch:02d}}.h5'
+                                                              .format(dataset_type=dataset_type)), verbose=1)
```
hgaiser: I know this is what PEP8 suggests, but I like the alternative more. Ie. : 
```python
foo(
    bar=1,
    baz=2
)
```

Could you change it to that format? The rest of the code is formatted the same way.

awilliamson:
@hgaiser Changes made. I took the liberty of liberally applying your style to the `ModelCheckpoint` as it could be cleaned further. Hopefully this meets your approval ( my editor did not want to cooperate on those PEP8 line constraints )

hgaiser:
Looks good to me, thanks!

yhenon-nextdroid:
Is this `./snapshots` safe for Windows users?

yhenon:
Is this `./snapshots` safe for Windows users?

de-vri-es:
Removing the `./` should have the same effect without any possible failure on Windows if the path is used properly. I do like the verbosity of `./snapshots`, but I suppose not breaking windows is more important. It might still work though... I'm never sure when forward slashes will actually break on windows and when not...

awilliamson:
I'm unsure if `./` would be safe on windows. Was just a default action on my part. Isn't this parsed via some other functions rather than direct access to os? I imagine that wrapper will sanitise it? Does a windows user want to check this for us?

Issue #68, opened 2017-11-17
Speed up augmentation of bounding boxes for non-trivial transforms


de-vri-es:
Bounding boxes are augmented in a rather expensive way right now for transforms other than flips. This can be improved.

wassname:
I noticed the [imgaug](https://github.com/aleju/imgaug) library that can transform masks and annotation points together. [Example](http://imgaug.readthedocs.io/en/latest/source/examples_bounding_boxes.html).

Tracyxzh:
```
# create image data generator objects
train_image_data_generator = keras.preprocessing.image.ImageDataGenerator(
        horizontal_flip=True,
    )
```
  These code is from train_csv.py . And you means there can have parameter like rotation_range?

de-vri-es:
A small status update:

Progress on the `cheap-augmentation` branch is quite far. All use of `keras.preprocessing.image.ImageDataGenerator` has either been removed (if it was a NOP) or replaced with the new `random_transform_generator(...)`. Applying the transformation is now the responsibility of `keras_retinanet.preprocessing.Generator`.

Unit tests for new code are still missing.

The branch has been rebased on `test-training` since that is probably going to be merged soon (and otherwise there will be conflicts).

de-vri-es:
Closing this in favour of #190.

liminghuiv:
Hi @de-vri-es, @Tracyxzh, @wassname,
Do you have any example on using imgaug to add addtional augmentation to keras-retinanet?

Thanks a lot.

Liming

Issue #186, opened 2017-12-15
when I used the ImageDataGenerator,I encountered a error


hnsywangxin:
The parameter of ImageDataGenerator is :
```
       horizontal_flip=False,
       featurewise_center=True,
        featurewise_std_normalization=True,
        rotation_range=10,
        width_shift_range=0.1,
        height_shift_range=0.1
```

and the error is :min() arg is an empty sequence.
the error is from "keras_retinanet/utils/image.py", line 80, in random_transform
    boxes[index, 0] = float(min(j))"
thank you

LaurensHagendoorn:
Jup, this can happen lot's basically a bounding box is augmented out of the image, and it cannot find it anymore, either lower your augmentation, or do something nasty like this: (or both)

`    while not finished:
        boxes=boxes_in
        if initseed is None:
            seed = np.random.randint(10000)
        else:
            seed = initseed

        image = image_data_generator.random_transform(image_in, seed=seed)

        # set fill mode so that masks are not enlarged
        fill_mode = image_data_generator.fill_mode
        image_data_generator.fill_mode = 'constant'

        for index in range(boxes_in.shape[0]):
            # generate box mask and randomly transform it
            mask = np.zeros_like(image, dtype=np.uint8)
            b = boxes_in[index, :4].astype(int)

            assert(b[0] < b[2] and b[1] < b[3]), 'Annotations contain invalid box: {}'.format(b)
            assert(b[2] <= image.shape[1] and b[3] <= image.shape[0]), 'Annotation ({}) is outside of image shape ({}).'.format(b, image.shape)

            mask[b[1]:b[3], b[0]:b[2], :] = 255

            mask_after = image_data_generator.random_transform(mask, seed=seed)[..., 0]
            mask_after = mask_after.copy()  # to force contiguous arrays

            # find bounding box again in augmented image
            [i, j] = np.where(mask_after == 255)
            if len(i)==0 or len(j)==0:
                print('Invalid augementation')
                wrongBox = True
                assert False
                continue
            boxes[index, 0] = float(min(j))
            boxes[index, 1] = float(min(i))
            boxes[index, 2] = float(max(j)) + 1 # set box to an open interval [min, max)
            boxes[index, 3] = float(max(i)) + 1 # set box to an open interval [min, max)

        if wrongBox:
            print('==== found a wrongbox! ====')
            wrongBox = False
            continue
        finished = True`

hgaiser:
Does this solve your issue?

[...]

Issue #181, opened 2017-12-13
Concern around random number seeding


mhwilder:
I ran into an issue while trying to implement class balancing using the np.random number generator. Basically, I was seeing a weird repeating sequence in the number generation and it was causing poor sampling. I think the root of the cause comes from line 58 in utils/image.py. Here np.random.randint is used to get a seed. Then when the Keras method image_data_generator.random_transform gets called, it calls np.random.seed(seed). In doing this, it biases what seed will be selected next. The following code snippet illustrates this problem:

```
import numpy as np
n = 10000
selected = np.zeros(n)
for i in range(100000):
    seed = np.random.randint(n)
    selected[seed] += 1
    np.random.seed(seed)
print('%d of %d selected' % (np.sum(selected>0), n))
print('Mean count of those selected is %0.1f' % np.mean(selected[selected>0]))
```

This is probably not causing many problems in the existing codebase because the python random library is used in most other places (using that library for my class balancing code also seemed to solve my issue). Additionally, not much augmentation is done in random_transform so repeating seeds may not be much of an issue. Still, it's probably worth considering alternatives for this because it could be a gotcha down the line as it was with me. I'm not sure what the best solution would be though.

As a side note, I wonder if in preprocessing/generator.py we should call random.seed(seed) in addition to np.random.seed(seed) since the goal is to get replicable sequences and random is used for the sequencing.

de-vri-es:
Using a global pseudo random number generator (plain python or numpy) is not really nice in the first place. Different code using the same PRNG may have different expectations of when it is (re)seeded.

The best solution (I think) is to have a unique PRNG with its own state for each thing that needs one. That would allow anyone to seed their own PRNG for reproducible results without interfering with other code. I'm not sure how easy that would be to do though.

de-vri-es:
Looks like (atleast for numpy), this would be a good solution: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.random.RandomState.html

I would recommend that you use this for your own algorithm so that you're not affected by others messing with the global PRNG.

We should also switch to a similar approach, but that requires rewriting the random augmentation code. That should be done together with #68 and #150.

mhwilder:
Thanks for the suggestion. It seems that RandomState would be a good solution for me and would be good to tie in elsewhere.

de-vri-es:
[...]

PR #190, opened 2017-12-16
Add random_transform_generator for augmenting images.

de-vri-es:
This PR adds a `random_transform_generator` to be used for augmenting data. It replaces the use of `keras.preprocessing.image.ImageDataGenerator`.

The main reason for doing this is that `ImageDataGenerator` creates a random transformation internally and applies it directly to the image. That doesn't give us the chance to apply the same transformation to bounding boxes.

The current workaround is to seed the global PRNG used by `ImageDataGenerator` with a random value and then have it transform first the image data, and then a mask for the bounding box (with the PRNG reset to the same random seed). The new bounding box is then determined from the transformed mask.

This is a rather expensive process, but also slightly fragile to chances in the `ImageDataGenerator`. It relies on the fact that `ImageDataGenerator` uses the global numpy PRNG, and that the number and order of generating random numbers is the same for the original image and the mask.

In contrast, the new `random_transform_generator` doesn't apply any transformations, it just generates them. The responsibility for applying the transforms to the image and bounding box has been moved into our own `Generator` with this PR.

`random_transform_generator` supports the following transformations:
* rotation
* translation
* scaling
* shearing
* flips

Notably, the one thing from `ImageDataGenerator.random_transform` it does not support is channel shifting, since that is not a homogeneous linear transformation in 2D.


keras_retinanet/preprocessing/generator.py
```diff
@@ -112,19 +116,32 @@ def resize_image(self, image):
...
+        if self.transform_generator is not None:
+            transform = adjust_transform_for_image(next(self.transform_generator), image)
+            apply_transform(transform, image, self.transform_parameters)
 
-            # randomly transform both image and annotations
-            image, annotations = random_transform(image, annotations, self.image_data_generator)
+            # Transform the bounding boxes in the annotations.
+            annotations = annotations.copy()
+            for index in range(annotations.shape[0]):
+                annotations[index, :4] = transform_aabb(transform, annotations[index, :4])
```
de-vri-es: One thing I was wondering: What should the generator do if `transform_aabb` puts the annotation outside of the image canvas? I'm assuming they will automatically be filtered now by `compute_input_output`. But maybe I overlooked something?

de-vri-es:
Note: should fix #68, #181 and I think #186.

de-vri-es:
Note: due to the holiday period, reviews are going to be slightly delayed. That also goes for this PR.

keras_retinanet/preprocessing/generator.py
de-vri-es: Also, is that desirable? When the aabb ends up outside the image and it is filtered, that means that part of the image becomes background, even though it was actually an annotation.

Should we instead discard the entire image? Or should we try up to X times to get a transformation that doesn't put any annotations outside of the image (and discard if that failed X times)? Or should we clip the new bounding box with the image canvas?

keras_retinanet/preprocessing/generator.py
```diff
@@ -112,19 +116,32 @@ def resize_image(self, image):
...
-    def preprocess_group(self, image_group, annotations_group):
-        for index, (image, annotations) in enumerate(zip(image_group, annotations_group)):
-            # preprocess the image (subtract imagenet mean)
-            image = self.preprocess_image(image)
+    def preprocess_group_entry(self, image, annotations):
+        # preprocess the image
+        image = self.preprocess_image(image)
+
+        # randomly transform both image and annotations
+        if self.transform_generator is not None:
```
hgaiser: Why not use `if self.transform_generator:` ?

keras_retinanet/preprocessing/generator.py
hgaiser: My first thought is to clip the annotation to the image canvas, to preserve as much of the annotations as possible. What do you think ?

keras_retinanet/utils/transform.py
```diff
@@ -0,0 +1,241 @@
...
+    ])
+
+    # Extract the min and max corners again.
+    min_corner = points.min(axis=1)
+    max_corner = points.max(axis=1)
+
+    return [min_corner[0], min_corner[1], max_corner[0], max_corner[1]]
+
+
+def _random_vector(min, max, prng = DEFAULT_PRNG):
```
hgaiser: Non-aligned default arguments are written down without spaces around the equal sign (same for some other places).

keras_retinanet/utils/transform.py
```diff
@@ -0,0 +1,241 @@
...
+        the translation matrix as 3 by 3 numpy array
+    """
+    return np.array([
+        [1, 0, translation[0]],
+        [0, 1, translation[1]],
+        [0, 0, 1]
+    ])
+
+
+def random_translation(min, max, prng = DEFAULT_PRNG):
```
hgaiser: Do we want to distinguish between X/Y translation min/max?

keras_retinanet/utils/transform.py
```diff
@@ -0,0 +1,241 @@
...
+    # Returns
+        a homogeneous 3 by 3 translation matrix
+    """
+    return translation(_random_vector(min, max, prng))
+
+
+def shear(amount):
+    """ Construct a homogeneous 2D shear matrix.
+    # Arguments
+        amount: the shear amount
```
hgaiser: Maybe make it a bit more clear what it is. Presumably it is the angle of the shear in radians?

keras_retinanet/utils/transform.py
```diff
@@ -0,0 +1,241 @@
...
+        [1, -np.sin(amount), 0],
+        [0,  np.cos(amount), 0],
+        [0, 0, 1]
+    ])
+
+
+def random_shear(min, max, prng = DEFAULT_PRNG):
+    """ Construct a random 2D shear matrix with shear angle between -max and max.
+    # Arguments
+        min:  the minumum shear factor.
```
hgaiser: Same here.

keras_retinanet/utils/transform.py
```diff
@@ -0,0 +1,241 @@
...
+        the zoom matrix as 3 by 3 numpy array
+    """
+    return np.array([
+        [factor[0], 0, 0],
+        [0, factor[1], 0],
+        [0, 0, 1]
+    ])
+
+
+def random_scaling(min, max, prng = DEFAULT_PRNG):
```
hgaiser: Same here, do we want to distinguish between x/y ?

keras_retinanet/utils/transform.py
philvdm: pep8 could be used for catching those spaces

de-vri-es: One problem with using `pytest --pep8` for that is that we do use spaces around the parameter assignment on multi-line function calls for alignment. The pep8 test doesn't distinguish between that and this type of whitespace.

keras_retinanet/utils/transform.py
de-vri-es: At this point I would say yes. These are generic transform functions. My initial reasoning was that you might want to scale the max and min with the image width and height. That isn't necessary now since the data generator does the scaling itself after generating the transform.

Still, I think it's a valid use case for a generic transform function.

keras_retinanet/utils/transform.py
de-vri-es: It is an angle, though the matrix generated isn't actually a shear matrix. However, I copied it from `keras` since I figured we should probably use the same definition of shear as `keras` does, even if it is technically incorrect.

I believe the `keras` definition of shear compensates for the stretching that a real shear operation would do by reducing the Y component as the X component gets larger. So it's sort of halfway between a shear and a rotation.

Another thing I wasn't really happy about is that this shear functions only shears parallel to the X axis. It's not very generic.

keras_retinanet/preprocessing/generator.py
de-vri-es: No reason really. Do you prefer it like that?

keras_retinanet/preprocessing/generator.py
de-vri-es: I'm not sure. I have a slight feeling that it might reduce the quality of training in some cases. Maybe some very distinctive part of the thing to recognize is clipped off, or maybe the clipped thing looks more like a different object (contrived example: a face with a clown nose vs a face with a normal nose).

keras_retinanet/utils/transform.py
de-vri-es: Anyway, fixed, I think.

keras_retinanet/utils/transform.py
de-vri-es: Damn you github, this comment thread is not outdated -.-

keras_retinanet/preprocessing/generator.py
hgaiser: Yeah.

keras_retinanet/utils/transform.py
hgaiser: But this code handles X and Y the same way right? Using the same min/max?

de-vri-es: no, min and max are 2D vectors (or a list/tuple/sequence of two elements)

hgaiser: Ah I see, I missed that part then. :+1:

keras_retinanet/preprocessing/generator.py
hgaiser: I think filtering it out entirely is more harmful then clipping. So either we make sure it doesn't occur, or we clip it in my opinion.

hgaiser:
As far as I can tell we only need to decide on what to do when an annotation is outside of the image canvas after transformation. When does this happen? Shouldn't the annotation always remain in the image canvas (aside from some rounding errors perhaps)?

de-vri-es:
No, rotations can easily put annotations completely or partly out of the canvas. The same is true for shearing and translation. Only flips don't have this problem, really.

yhenon:
Two things:
- Currently bounding boxes that overlap an anchor with 0.4 < IoU < 0.5 are tagges as `don't care` and do not contribute to the loss. It might make sense to consider doing something similar for boxes that are clipped by a transformation.
- whatever the ultimate decision, we should let the threshold be user configurable.

keras_retinanet/preprocessing/generator.py
de-vri-es: Updated.

hgaiser:
Shall we move on to merging this, and worry about that edge case in a later PR?

LaurensHagendoorn:
If by the edge case you mean when bounding boxes end up outside the image, I think it should be fixed first. I have been getting that result with the current augmentations quite a lot, and it is a problem. I suppose clipping when possible, and not crashing otherwise would be my preferred solution.

hgaiser:
Crashing is not supposed to happen. I propose to indeed either clip for now, or leave it untouched. Personally I still believe clipping would be fine.

LaurensHagendoorn:
Yeah I think crashing was actually fixed by a change a while back. I am in favour of merging this soon though! I have been oogling this branch for a while now, would be awesome to be able to use this!

de-vri-es:
I also don't think it will crash now. I believe it will generate invalid annotations which will be filtered out automatically. For my part, we can make an issue to resolve the discarding/clipping problem and merge as is.

de-vri-es:
I tagged a 0.1 pre-release, since we seemed to be pretty stable.

Rebased this branch on master. Once CI passes I'll merge.

de-vri-es:
Discussion on what to do with the annotations partially transformed out of the canvas can continue at #223

PR #196, opened 2017-12-19
VOC 2007 Evaluation and Visualisation

awilliamson:
Evaluation of VOC 2007 data for mAP, including bounding box visualisation for ground truth and detections.

Class based due to all the arrays, preprocessing, state etc required for this evaluation. Facilitated visualisation and saving too.

### DIFF
--- a/keras_retinanet/utils/voc_eval.py
+++ b/keras_retinanet/utils/voc_eval.py
@@ -56,20 +56,12 @@ class VOCEvaluator():
         return ap
 
     @staticmethod
-    def draw_gt_bboxes(image, anno_list):
-        for anno in anno_list:
-            label = anno['category_id']
-            b = np.array(anno['bbox']).astype(int)
-
-            cv2.rectangle(image, (b[0], b[1]), (b[2], b[3]), (0, 255, 0), 1)
-            caption = "GT:{}".format(label)
-            cv2.putText(image, caption, (b[0], b[1] - 10), cv2.FONT_HERSHEY_PLAIN, 1, (0, 0, 0), 3)
-            cv2.putText(image, caption, (b[0], b[1] - 10), cv2.FONT_HERSHEY_PLAIN, 1, (255, 255, 255), 2)
-
-    @staticmethod
-    def filter_top_N_bboxes(bbox_preds, N):
-        max_liklihood_sorted = sorted(bbox_preds, key=lambda x: x['score'], reverse=True)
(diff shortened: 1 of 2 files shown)


de-vri-es:
Thanks for the PR. Just a short heads-up: due to the holiday period, reviews are probably going to be delayed a bit. My own PR is also on hold until my colleagues get back from slacking off with family and friends ;)

hgaiser:
Pffff

awilliamson:
No worries at all :) I understand completely. Feel free to take your time. I'd rather these things be thoroughly reviewed and progress the retinanet implementation correctly than hastily accepted. Most of this work stems from my own necessity and my own research project needs. Just a strong believer in open source.

de-vri-es:
Just a quick note (not a formal review yet), I think we'll want a script that can do evaluation for all the available generators (so also COCO and CSV). There is something for COCO already (although I'm not sure how feature complete that is), but ideally we'll have one script like `train.py` but then for evaluation.

If that is something you would like to work on, that would be awesome. Otherwise, we could take your PR and merge it into one evaluation script afterwards.

awilliamson:
Of course, abstracting evaluation mechanisms is a good end-goal, providing extensibility; however for now at least some measure of functionality should be added, then iterated upon.
I wouldn't mind working on a centralised system for Evaluators, considering I wrote some of the class backbone already. But when this would happen I am unsure.

hgaiser:
Indeed, as @de-vri-es mentioned, preferably we have something that is dataset-agnostic.

What is the difference between this PR and the work that was done in https://github.com/fizyr/keras-retinanet/tree/evaluation ? That branch is a bit outdated and the work is not tested properly yet, but it seems to be doing more or less the same thing (except for saving visualized results).

awilliamson:
@hgaiser Indeed, it is more or less the same as the evaluation metric is derived directly from faster-rcnn. It's written up in a more class oriented way which will enable and facilitate further development towards and dataset agnostic system, or at least an abstract system similar to how Generators work.

Work on command-line args for saving paths and overlaying. Overall a more feature rich evaluator, useful for interpreting the mAP score reported visually.

I completely agree with the aim of the more abstract system, but currently time is limited for me due to deliverable deadlines approaching. I'd consider the class based VOCEvaluator as a good initial step towards that goal.

hgaiser:
Merged into `evaluation` branch so that I can continue this work from there, thanks!

PR #210, opened 2018-01-04
Refactor of ResNet model generation.

hgaiser:
This PR removes some duplicate code for generating resnet models.

Also, the name of the function was changed from `ResNet50RetinaNet` to `resnet50_retinanet`.

### DIFF
--- a/keras_retinanet/models/resnet.py
+++ b/keras_retinanet/models/resnet.py
@@ -19,66 +19,67 @@ import keras_resnet
 import keras_resnet.models
 from ..models import retinanet
 
-WEIGHTS_PATH_NO_TOP_50 = 'https://github.com/fizyr/keras-models/releases/download/v0.0.1/ResNet-50-model.keras.h5'
-WEIGHTS_PATH_NO_TOP_101 = 'https://github.com/fizyr/keras-models/releases/download/v0.0.1/ResNet-101-model.keras.h5'
-WEIGHTS_PATH_NO_TOP_152 = 'https://github.com/fizyr/keras-models/releases/download/v0.0.1/ResNet-152-model.keras.h5'
+resnet_filename = 'ResNet-{}-model.keras.h5'
+resnet_resource = 'https://github.com/fizyr/keras-models/releases/download/v0.0.1/{}'.format(resnet_filename)
 
 custom_objects = retinanet.custom_objects.copy()
 custom_objects.update(keras_resnet.custom_objects)
 
 
-def ResNet50RetinaNet(inputs, num_classes, weights='imagenet', *args, **kwargs):
-    image = inputs
+def resnet_retinanet(num_classes, backbone='50', inputs=None, weights='imagenet', **kwargs):
+    allowed_backbones = ['50', '101', '152']
+    if backbone not in allowed_backbones:
+        raise ValueError('Backbone (\'{}\') not in allowed backbones ({}).'.format(backbone, allowed_backbones))
 
-    # load pretrained imagenet weights?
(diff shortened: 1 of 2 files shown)


keras_retinanet/models/resnet.py
```diff
@@ -19,66 +19,69 @@
...
+resnet_filename = 'ResNet-{}-model.keras.h5'
+resnet_resource = 'https://github.com/fizyr/keras-models/releases/download/v0.0.1/{}'.format(resnet_filename)
 
 custom_objects = retinanet.custom_objects.copy()
 custom_objects.update(keras_resnet.custom_objects)
 
 
-def ResNet50RetinaNet(inputs, num_classes, weights='imagenet', *args, **kwargs):
-    image = inputs
+def resnet_retinanet(num_classes, backbone='50', inputs=None, weights='imagenet', **kwargs):
```
de-vri-es: `backbone` as string seems a bit weird. Why not just use a the numbers 50, 101 and 152?

keras_retinanet/models/resnet.py
```diff
@@ -19,66 +19,69 @@
...
+    if backbone not in allowed_backbones:
+        raise ValueError('Backbone (\'{}\') not in allowed backbones ({}).'.format(backbone, allowed_backbones))
 
-    # load pretrained imagenet weights?
+    # choose default input
+    if inputs is None:
+        inputs = keras.layers.Input(shape=(None, None, 3))
+
+    # determine which weights to load
     if weights == 'imagenet':
```
de-vri-es: I suppose this isn't new behaviour, but this seems like a weird place to download the weights. It also means that you can't use a file called 'imagenet' as weights. Not a huge problem, but it is a bit dirty.

Personally I'd put the downloading in a separate function that a script can invoke before creating the model. That would leave this function a lot cleaner.

keras_retinanet/models/resnet.py
```diff
@@ -19,66 +19,69 @@
...
         weights_path = keras.applications.imagenet_utils.get_file(
-            'ResNet-50-model.keras.h5',
-            WEIGHTS_PATH_NO_TOP_50, cache_subdir='models', md5_hash='3e9f4e4f77bbe2c9bec13b53ee1c2319'
+            filename,
+            resource,
+            cache_subdir='models',
+            md5_hash=checksum
         )
+    elif weights == 'none':
+        weights_path = None
```
de-vri-es: I wouldn't allow `'none'` as string. Like `weights='imagenet'` it claims another valid filename as a special case. I'd force the user to pass a real `None` instead.

In my opinion, special handling for strings is something better done in argument parsing of a script than in API functions.

keras_retinanet/models/resnet.py
hgaiser: It used to be `'resnet50'` but I changed that later on to `'50'`. But you're right, I'll change it.

keras_retinanet/models/resnet.py
hgaiser: I like it :+1:

keras_retinanet/models/resnet.py
hgaiser: Fair enough.

hgaiser:
@de-vri-es I made the requested changes.

de-vri-es:
Looks good, thanks.

de-vri-es:
How's the backwards compatibility coming along? Are you still working on that?

hgaiser:
Working is a big word, but yeah. Awaiting tests now, then it can get merged.

Issue #227, opened 2018-01-16
data augmentation error


LinlyAC:
I found that you have updated about data augmentation.
I made this data augmentation in my code for finetuning a new model based on your best model Resnet50,

![1](https://user-images.githubusercontent.com/16618172/34981103-78086360-fae1-11e7-839f-dd9668a17fc3.png)

but the loss is werid:

![3](https://user-images.githubusercontent.com/16618172/34982207-bd19955c-fae4-11e7-9cc0-45ae37eedc50.png)

the classfcation loss is very high but the regression loss is almost to zero. Is there something wrong?

de-vri-es:
Rotation is in radians, so 40 radians is quite a few full circles. Translation is relative to the width/height of the image, so max translation is ten times the width/height of your images. That easily puts most of your images completely out of the image canvas. I expect that could very well explain bad training results.

Could you try with some lower values?

hgaiser:
Should we add a check for the translation so that it is between [0, 1) ? No sense in other values I suppose?

LinlyAC:
@de-vri-es so what you say is my rotation is small and the translation is too big?

de-vri-es:
> @de-vri-es so what you say is my rotation is small and the translation is too big?

Actually, they're both very big. I assume you meant 40 degrees which is only 0.698 radians. Unless you have a reason not to, I'd also center the rotation around 0, so set min to -0.35 and max to 0.35 or something along those lines.

> Should we add a check for the translation so that it is between [0, 1) ? No sense in other values I suppose?

The only problem with that is that it is the data generator that interprets the translation as factor of the image size. The `random_transform_generator`  truly puts values in the range of whatever you give. Limiting that to `[0, 1)` makes it much less generic.

We could also simply make the translation absolute in pixels, instead of interpreting it as factor of the image size. Or at the very least document it

LinlyAC:
thank you very much!

LinlyAC:
another question, if I use the new augmentation way, should I adjust some optimizer parameters, such as adam lr?

de-vri-es:
I submitted a PR (#228) to let the user choose absolute translations, and to improve the documentation. Hopefully that will make it more clear what values make sense here. 

@LinlyAC: Could you have a look at #228 if the updated documentation is clear enough according to you?

> another question, if I use the new augmentation way, should I adjust some optimizer parameters, such as adam lr?

[...]

PR #228, opened 2018-01-16
Relative or absolute translation

de-vri-es:
This PR allows translation to be interpreted as absolute pixels instead of a factor of the image size. Additionally, it tries to improve the documentation for `random_transform[_generator]` regarding units.

This should hopefully prevent misinterpretation of the values as in #227.

### DIFF
--- a/keras_retinanet/utils/transform.py
+++ b/keras_retinanet/utils/transform.py
@@ -205,6 +205,11 @@ def random_transform(
       * flip x (if applied)
       * flip y (if applied)
 
+    Note that by default, the data generators in `keras_retinanet.preprocessing.generators` interpret the translation
+    as factor of the image size. So an X translation of 0.1 would translate the image by 10% of it's width.
+    See `keras_retinanet.utils.image.TransformParameters` for a way to disable this behaviour and use absolute pixel
+    translation instead.
+
     # Arguments
         min_rotation:    The minimum rotation for the transform as scalar.
         max_rotation:    The maximum rotation for the transform as scalar.
--- a/keras_retinanet/utils/transform.py
+++ b/keras_retinanet/utils/transform.py
@@ -103,16 +103,16 @@ def random_translation(min, max, prng=DEFAULT_PRNG):
     return translation(_random_vector(min, max, prng))
 
 
-def shear(amount):
+def shear(angle):
     """ Construct a homogeneous 2D shear matrix.
     # Arguments
-        amount: the shear amount
+        angle: the shear angle in radians
     # Returns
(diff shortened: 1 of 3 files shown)


keras_retinanet/utils/transform.py
```diff
@@ -205,13 +205,18 @@ def random_transform(
       * flip x (if applied)
       * flip y (if applied)
 
+    Note that by default, the data generators in `keras_retinanet.preprocessing.generators` interpret the translation
+    as factor of the image size. So an X translation of 0.1 would translate the image by 10% of it's width.
+    See `keras_retinanet.utils.image.TransformParameters` for a way to disable this behaviour and use absolute pixel
```
hgaiser: Why not directly reference `relative_translation` here?

keras_retinanet/utils/transform.py
```diff
@@ -228,9 +233,35 @@ def random_transform(
...
+      * rotation
+      * translation
+      * shear
+      * scaling
+      * flip x (if applied)
+      * flip y (if applied)
+
+    Note that by default, the data generators in `keras_retinanet.preprocessing.generators` interpret the translation
+    as factor of the image size. So an X translation of 0.1 would translate the image by 10% of it's width.
+    See `keras_retinanet.utils.image.TransformParameters` for a way to disable this behaviour and use absolute pixel
```
hgaiser: Here too

keras_retinanet/utils/transform.py
de-vri-es: Good point, will change. When I wrote that I hadn't actually implemented it yet, so I didn't know what it was called yet :)

hgaiser: Ah I see, well after that it LGTM.

keras_retinanet/utils/transform.py
```diff
@@ -205,13 +205,18 @@ def random_transform(
       * flip x (if applied)
       * flip y (if applied)
 
+    Note that by default, the data generators in `keras_retinanet.preprocessing.generators` interpret the translation
+    as factor of the image size. So an X translation of 0.1 would translate the image by 10% of it's width.
+    Set `relative_translation` to `False` in the `TransformParameters` of a data generator to have it interpret
+    the translation directly as pixel distances instead.
```
de-vri-es: @hgaiser: Tweaked it to this. What do you think?

hgaiser: Looks good to me!

de-vri-es: Alright, will merge after CI passes =]

Issue #225, opened 2018-01-15
validation in train.py 'csv' 


Weixing-Zhang:
In keras-retinanet/keras_retinanet/bin/train.py (branch: master), I found Line 206 (csv_parser.add_argument('--val-annotations' ... ) won't be executed if the dataset_type isn't 'coco' (Line 97-106). Am I missing something here?

de-vri-es:
I think it should work as intended:

https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/bin/train.py#L201

https://github.com/fizyr/keras-retinanet/blob/8f22f359949cf8cbf0886adaa61abd84fcd24d89/keras_retinanet/bin/train.py#L145-L160

Or am I the one missing something ?

Weixing-Zhang:
@de-vri-es 
That's right. train.py creates generator for validation. But I meant was, in callbacks = create_callbacks (Line 246), passed validation_generator only happens in the if statement (Line 97-103).

de-vri-es:
hmm, yeah, it looks like you're right. I'm not sure that there is a proper callback we can use for evaluating with a CSV generator though.

de-vri-es:
I don't really know. I'd recommend just looking at the COCO dataset and see from there.

I do know that @hgaiser is working on an evaluation script on the `evaluation` branch. Maybe work from there can be used to make a validation callback for the training process as well. Ideally one that doesn't care which data generator is used.

hgaiser:
As @de-vri-es said, there isn't a callback for anything other than COCO at the moment. Indeed the `evaluation` branch could (should?) include a callback for any dataset type.

For now I'll consider this issue resolved.

Weixing-Zhang:
Thanks you all!

de-vri-es:
For the archives: this should now be fixed by #231.

PR #231, opened 2018-01-18
Evaluation tool

hgaiser:
This PR merges previous work from https://github.com/fizyr/keras-retinanet/pull/196 and previous work done on the `evaluation` branch. This tool should provide `mAP` evaluation scores for arbitrary generators (and therefore datasets).

The metric computed is largely derived from [`py-faster-rcnn`](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/datasets/voc_eval.py). Compared to the COCO evaluation it should theoretically the same as AP@0.5. Testing on the pretrained COCO model gives an `mAP` of 0.507, whereas the COCO evaluation gives an `AP@0.5` of 0.513. I'm not sure where this discrepancy comes from, this is something that could be further investigated in the future.

@awilliamson could you have a look too ?


keras_retinanet/utils/eval.py
```diff
@@ -0,0 +1,192 @@
...
+        # find the order with which to sort the scores
+        scores_sort = np.argsort(-scores)[:max_detections]
+
+        # select detections
+        image_boxes      = detections[0, indices[0][scores_sort], :4]
+        image_scores     = np.expand_dims(detections[0, indices[0][scores_sort], 4 + indices[1][scores_sort]], axis=1)
+        image_detections = np.append(image_boxes, image_scores, axis=1)
+        image_predicted_labels = indices[1][scores_sort]
+
+        if save_path:
```
de-vri-es: I would go for `if save_path is not None:` here. Otherwise, `--save-path ''` on the command line doesn't save any images. I would expect the images to be saved to the current working directory. I would be a bit annoyed if it turns out afterwards that no images were saved.

`--save-path .` would be a workaround, but I don't see a reason to special case empty strings here.

de-vri-es:
Code looks good in general. I didn't test it manually, I do think we should do that before merging (if you haven't already). I do have a few remarks (besides the in-line comment):

* There is no documentation for the added functions. Even for internal functions, some documentation on the intended behaviour and structure of data would be very useful for future maintenance.

* There are quite some command line options in common with the train script. I'm not sure that we should share these between the scripts, but if not we should manually make sure they stay in sync with each other in the future.

* It would also be nice to be able to specify a random transform generator on the command line to see the performance of the network on augmented data (and to test the random transform generator ^_^). We can add this in a separate PR to both training and testing scripts though.

* There doesn't seem to be an evaluation callback for use with training as discussed in #225. It can be a separate PR, but if nobody is working on it right now I'd like to re-open the issue so that we don't forget it.

keras_retinanet/utils/eval.py
hgaiser: Agreed, edited.

hgaiser:
Thanks for the review. Fair point on the documentation, I will add it soon. Also thanks for reminding me of the callback, should be fairly straightforward to add that. I'll let you know when I'm done updating.

hgaiser:
Alright I updated it with documentation and a callback. Can you review @de-vri-es ?

keras_retinanet/bin/train.py
```diff
@@ -208,7 +210,8 @@ def parse_args(args):
     parser.add_argument('--steps',         help='Number of steps per epoch.', type=int, default=10000)
     parser.add_argument('--snapshot-path', help='Path to store snapshots of models during training (defaults to \'./snapshots\')', default='./snapshots')
     parser.add_argument('--no-snapshots',  help='Disable saving snapshots.', dest='snapshots', action='store_false')
-    parser.set_defaults(snapshots=True)
+    parser.add_argument('--no-evaluation', help='Disable per epoch evaluation.', dest='evaluation', action='store_false')
+    parser.set_defaults(snapshots=True, evaluation=True)
```
de-vri-es: Same as above: the `set_defaults` is not needed.

keras_retinanet/bin/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+    csv_parser.add_argument('classes', help='Path to a CSV file containing class label mapping.')
+
+    parser.add_argument('model',             help='Path to RetinaNet model.')
+    parser.add_argument('--gpu',             help='Id of the GPU to use (as reported by nvidia-smi).')
+    parser.add_argument('--score-threshold', help='Threshold on score to filter detections with (defaults to 0.05).', default=0.05, type=float)
+    parser.add_argument('--iou-threshold',   help='IoU Threshold to count for a positive detection (defaults to 0.5).', default=0.5, type=float)
+    parser.add_argument('--max-detections',  help='Max Detections per image (defaults to 100).', default=100, type=int)
+    parser.add_argument('--save-path',       help='Path for saving images with detections.')
+
+    parser.set_defaults(snapshots=True)
```
de-vri-es: The `set_defaults` is not needed. `store_true` and `store_false` already set the default value to the opposite value: https://docs.python.org/3/library/argparse.html#action

hgaiser:
Ah I see, I didn't know that. I updated it now, shall I merge it then?

de-vri-es:
Yeah, good with me =]

Weixing-Zhang:
@de-vri-es 
Thanks a lot! Please allow me to have you in my paper's acknowledgements as same as Hans Gaiser. Maarten de Vries? Let me know if I misspell your name.

de-vri-es:
That is correct. Thanks :)

PR #232, opened 2018-01-18
Add skip_mismatch to weight loading.

hgaiser:
This PR adds the `skip_mismatch` flag to weights loading. This flag requires Keras 2.1.3, so that version is also bumped. Alternatively I could change this such that it doesn't use `skip_mismatch` if Keras 2.1.2 is found, but I prefer to bump the minimum version instead.

### DIFF
--- a/keras_retinanet/models/resnet.py
+++ b/keras_retinanet/models/resnet.py
@@ -50,7 +50,7 @@ def download_imagenet(backbone):
     )
 
 
-def resnet_retinanet(num_classes, backbone=50, inputs=None, weights='imagenet', **kwargs):
+def resnet_retinanet(num_classes, backbone=50, inputs=None, weights='imagenet', skip_mismatch=True, **kwargs):
     allowed_backbones = [50, 101, 152]
     if backbone not in allowed_backbones:
         raise ValueError('Backbone (\'{}\') not in allowed backbones ({}).'.format(backbone, allowed_backbones))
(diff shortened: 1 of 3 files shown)


keras_retinanet/models/resnet.py
```diff
@@ -80,7 +80,7 @@ def resnet_retinanet(num_classes, backbone=50, inputs=None, weights='imagenet',
 
     # optionally load weights
     if weights_path:
-        model.load_weights(weights_path, by_name=True)
+        model.load_weights(weights_path, by_name=True, skip_mismatch=True)
```
de-vri-es: Shouldn't we leave this behaviour overridable? We can default it to true, but if you're sure that the weights match the model (for example because you just trained it), not skipping mismatched weights does give some additional error checking.

hgaiser: Hmm you raise a valid point. The most common case would be, I expect, that you would intentionally use a mismatching model (for example by always starting with the pretrained COCO model when training a custom dataset). Therefore the default in my opinion should at least be `skip_mismatch=True`. I'm not sure if it then makes sense to allow an option to set `skip_mismatch=False`, but there's no harm in adding it.

de-vri-es: > The most common case would be, I expect, that you would intentionally use a mismatching model (for example by always starting with the pretrained COCO model when training a custom dataset).

Is this function also used when testing/evaluating? Or is it more common to load the model and weights directly from a file without building the model first?

hgaiser: > Is this function also used when testing/evaluating? Or is it more common to load the model and weights directly from a file without building the model first?

It can be used for testing / evaluation, but I would advise against it. Sometimes it can prove useful, for example if there was a small bug in the architecture, with which a network was trained. You can then create a new model with a fixed architecture and load in the weights using this function.

Overall, I would say this function should only be used for training.

de-vri-es: Right, then I agree defaulting to true makes sense.

PR #241, opened 2018-01-24
Allow resuming of training via snapshots.

hgaiser:
This PR allows to resume training from a snapshot. The difference with resuming and loading in weights, is that snapshots also restores the optimizer state.

An unfortunate requirement is that the snapshot is created with this PR and that multi-gpu is disabled. I'm looking into a fix for this, if I find a suitable one it will be submitted as a new PR.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -38,7 +38,7 @@ from ..callbacks import RedirectModel
 from ..callbacks.eval import Evaluate
 from ..preprocessing.pascal_voc import PascalVocGenerator
 from ..preprocessing.csv_generator import CSVGenerator
-from ..models.resnet import resnet50_retinanet
+from ..models.resnet import resnet50_retinanet, custom_objects
 from ..utils.transform import random_transform_generator
 from ..utils.keras_version import check_keras_version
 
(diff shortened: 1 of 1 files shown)


keras_retinanet/bin/train.py
```diff
@@ -202,7 +208,8 @@ def parse_args(args):
     csv_parser.add_argument('classes', help='Path to a CSV file containing class label mapping.')
     csv_parser.add_argument('--val-annotations', help='Path to CSV file containing annotations for validation (optional).')
 
-    parser.add_argument('--weights',       help='Weights to use for initialization (defaults to ImageNet).', default='imagenet')
+    parser.add_argument('--weights',       help='Weights to use for initialization (defaults to \'imagenet\').', default='imagenet')
+    parser.add_argument('--snapshot',      help='Snapshot to resume training with.')
```
de-vri-es: We could use a mutually exclusive group for `--weights` and `--snapshot`: https://docs.python.org/3/library/argparse.html#mutual-exclusion

hgaiser: Perfect! Added it.

mrtgenet:
Hello people !
My question may seem naive (I didn't find much about it in other issues) but is there a way to force epoch number to be the one from the loaded snapshot ? 
I mean, for example training failed at epoch 11, whatever the reason (hardware, ctrl-c...). So we can re-start training using '--snapshot ./snapshots/modelname_csv_10.h5' which is great, but yet the resumed training starts at epoch 1 (with the effective epoch 10 model losses and stuff) instead of 11. So I know new epoch 1 is actually epoch 11 but it's perturbing at first, and it means new snapshots will erase old ones ?

Thanks for any remark, and sorry if it's a bit confused :D

PR #243, opened 2018-01-26
Add option to freeze backbone layers.

hgaiser:
This PR adds the option to freeze the backbone layers (ie., stop them from updating further during training).

This is often useful when doing transfer learning.

### DIFF
--- a/keras_retinanet/models/retinanet.py
+++ b/keras_retinanet/models/retinanet.py
@@ -182,10 +182,16 @@ def retinanet(
     anchor_parameters       = AnchorParameters.default,
     create_pyramid_features = __create_pyramid_features,
     submodels               = None,
+    freeze_backbone         = False,
     name                    = 'retinanet'
 ):
     if submodels is None:
         submodels = default_submodels(num_classes, anchor_parameters)
+
+    # when doing transfer learning, freeze the backbone layers
+    if freeze_backbone:
+        for l in backbone.layers:
+            l.trainable = False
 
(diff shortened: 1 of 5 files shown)


keras_retinanet/models/retinanet.py
```diff
@@ -181,11 +181,17 @@ def retinanet(
     anchor_parameters       = AnchorParameters.default,
     create_pyramid_features = __create_pyramid_features,
     submodels               = None,
+    freeze_backbone         = False,
```
de-vri-es: I'm wondering if adding another parameter here is the smart thing to do. The added parameter doesn't make the function more flexible. The backbone could have been frozen before passing it to this function instead.

If it is currently to cumbersome to freeze the whole backbone, maybe we could add a utility function to freeze an entire model.

For the `resnet` functions we could add callback to modify the backend model before passing it down to `retinanet` rather than specifically adding a `freeze_backend` option. That would be more future proof (allow more modifications without exploding the number of arguments).

hgaiser: I see your point. If I have the time for it I'll implement this, but it could be that I postpone this a bit.

de-vri-es:
Rebased and implemented freezing as generic modifier callback and a utility function.

hgaiser:
Shall we merge this @de-vri-es ?

de-vri-es:
I haven't tested it properly yet. Have you?

hgaiser:
Updated and tested, LGTM!

PR #246, opened 2018-01-27
Fix applying transforms to images.

de-vri-es:
There were some problems with applying transforms to images:

* The first axis of an image stored as numpy array is the Y axis, not the X axis.

* The underlying transform function from scipy interprets the given matrix as a transformation from the original canvas to the transformed image. That is somewhat counter-intuitive: it causes a scaling of (3, 3) to **shrink** the image contents by a factor 3.

* There also seems to be an off-by-one error in scipy's transform function.

This PR changes the transformation before applying it to the image to make sure that `utils.image.apply_transform` behaves as expected.

The PR also adds an option to keep looping over a dataset to `debug.py`. That was quite useful for looking at random transformations.

### DIFF
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -93,6 +93,7 @@ def parse_args(args):
     csv_parser.add_argument('annotations', help='Path to CSV file containing annotations for evaluation.')
     csv_parser.add_argument('classes',     help='Path to a CSV file containing class label mapping.')
 
+    parser.add_argument('-l', '--loop', help='Loop forever, even if the dataset is exhausted.', action='store_true')
     parser.add_argument('--no-resize', help='Disable image resizing.', dest='resize', action='store_false')
     parser.add_argument('--annotations', help='Show annotations on the image.', action='store_true')
     parser.add_argument('--random-transform', help='Randomly transform image and annotations.', action='store_true')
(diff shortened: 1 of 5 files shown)


de-vri-es:
> There also seems to be an off-by-one error in scipy's transform function.

Actually, the origin of linear transformations was at the **center** of the pixel at (0, 0), rather than the top-left corner of the image. Updated PR to fix that the right way.

hgaiser:
The code LGTM, I assume you've tested it?

de-vri-es:
Yes. Did you still want to disable non-flips?

hgaiser:
> Yes. Did you still want to disable non-flips?

Yeah, the transformed images don't look natural to me.

Example:

coco

It's probably the `scipy` affine transform function that does this. Until that is cleared up I don't think we should enable all transforms.

de-vri-es:
Considering that the `scipy` method is quite... imperfect, both in API, performance and end-result, I looked around for alternatives.

`cv2.warpAffine` seems like a good candidate. OpenCV already considers the X axis of an image to point right and the Y axis to point down. It also uses the more intuitive interpretation of the matrix, so that a scaling of 0.5 shrinks the image content rather than enlarges it. It can also work on a multichannel image in one go without splitting and later merging the channels. It's really better on every front...

I wouldn't be surprised if it is also significantly faster, so who knows, it might already improve #249. A few quick tests showed that rotated lines are much smoother, but a second opinion would be a good idea.

hgaiser:
Ahw yesh, much better :)

screenshot from 2018-01-31 20-27-29

As far as I'm concerned it can be merged then.

de-vri-es:
Alrighty then. This time I also did extensive testing, so I feel comfortable merging this.

PR #253, opened 2018-01-29
Adding Google Open Images Dataset generator

lvaleriu:
Let me bring my contribution to this nice project. I added a new generator who is able to load the https://github.com/openimages/dataset.

Since this dataset contains many classes (545 trainable box-level classes) i've added a filter option in order to choose a smaller number of classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,16 @@ The pretrained MS COCO model can be downloaded [here](https://github.com/fizyr/k
  Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.623
 ```
 
+### Usage
+For training on [OID](https://github.com/openimages/dataset), run:
+```shell
+# Running directly from the repository:
+keras_retinanet/bin/train.py oid /path/to/OID 2017_11
+
+# Using the installed script:
+retinanet-train pascal /path/to/OID train 2017_11
+```
+
 For training on a custom dataset, a CSV file can be used as a way to pass the data.
 See below for more details on the format of these CSV files.
 To train using your CSV, run:
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ retinanet-train oid /path/to/OID
 You can also specify a list of labels if you want to train on a subset
 by adding the argument 'labels_filter':
 
-keras_retinanet/bin/train.py oid /path/to/OID --labels_filter='['Helmet', 'Tree']'
+keras_retinanet/bin/train.py oid /path/to/OID --labels_filter=Helmet,Tree
 ```
 
 For training on a custom dataset, a CSV file can be used as a way to pass the data.
--- a/README.md
+++ b/README.md
@@ -66,7 +66,11 @@ For training on [OID](https://github.com/openimages/dataset), run:
 keras_retinanet/bin/train.py oid /path/to/OID 2017_11
 
 # Using the installed script:
-retinanet-train pascal oid /path/to/OID train 2017_11
+retinanet-train oid /path/to/OID 2017_11
+
+You can specify a list of labels if you want to train on a subset
+by adding the parameter labels_filter:
+    labels_filter = ["Helmet", "Tree", "Stop sign", "Traffic light"]
 ```
 
 For training on a custom dataset, a CSV file can be used as a way to pass the data.
--- a/README.md
+++ b/README.md
@@ -62,14 +62,15 @@ The pretrained MS COCO model can be downloaded [here](https://github.com/fizyr/k
 For training on [OID](https://github.com/openimages/dataset), run:
 ```shell
 # Running directly from the repository:
-keras_retinanet/bin/train.py oid /path/to/OID 2017_11
+keras_retinanet/bin/train.py oid /path/to/OID
 
 # Using the installed script:
-retinanet-train oid /path/to/OID 2017_11
+retinanet-train oid /path/to/OID
 
-You can specify a list of labels if you want to train on a subset
-by adding the parameter labels_filter:
-    labels_filter = ["Helmet", "Tree", "Stop sign", "Traffic light"]
+You can also specify a list of labels if you want to train on a subset
+by adding the argument 'labels_filter':
+
+keras_retinanet/bin/train.py oid /path/to/OID --labels_filter='['Helmet', 'Tree']'
 ```
 
 For training on a custom dataset, a CSV file can be used as a way to pass the data.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,6 @@ The pretrained MS COCO model can be downloaded [here](https://github.com/fizyr/k
  Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.623
 ```
 
-### Usage
 For training on [OID](https://github.com/openimages/dataset), run:
 ```shell
 # Running directly from the repository:
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ For training on [OID](https://github.com/openimages/dataset), run:
 keras_retinanet/bin/train.py oid /path/to/OID 2017_11
 
 # Using the installed script:
-retinanet-train pascal /path/to/OID train 2017_11
+retinanet-train pascal oid /path/to/OID train 2017_11
 ```
 
 For training on a custom dataset, a CSV file can be used as a way to pass the data.
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -224,10 +224,13 @@ def parse_args(args):
     pascal_parser = subparsers.add_parser('pascal')
     pascal_parser.add_argument('pascal_path', help='Path to dataset directory (ie. /tmp/VOCdevkit).')
 
+    def csv_list(string):
+        return string.split(',')
+
     oid_parser = subparsers.add_parser('oid')
     oid_parser.add_argument('main_dir', help='Path to dataset directory.')
     oid_parser.add_argument('--version',  help='The current dataset version is V3.', default='2017_11')
-    oid_parser.add_argument('--labels_filter',  help='A list of labels to filter.', default=None)
+    oid_parser.add_argument('--labels_filter',  help='A list of labels to filter.', type=csv_list, default=None)
     oid_parser.add_argument('--annotation_cache_dir', help='Path to store annotation cache.', default='.')
 
     csv_parser = subparsers.add_parser('csv')
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -170,20 +170,19 @@ def create_generators(args):
             subset='train',
             version=args.version,
             labels_filter=args.labels_filter,
+            annotation_cache_dir=args.annotation_cache_dir,
             transform_generator=transform_generator,
             batch_size=args.batch_size
         )
 
-        if args.val_annotations:
-            validation_generator = OpenImagesGenerator(
-                args.main_dir,
-                subset='validation',
-                version=args.version,
-                labels_filter=args.labels_filter,
-                batch_size=args.batch_size
-            )
-        else:
-            validation_generator = None
+        validation_generator = OpenImagesGenerator(
+            args.main_dir,
+            subset='validation',
+            version=args.version,
+            labels_filter=args.labels_filter,
+            annotation_cache_dir=args.annotation_cache_dir,
+            batch_size=args.batch_size
+        )
     else:
         raise ValueError('Invalid data type received: {}'.format(args.dataset_type))
 
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -168,6 +168,7 @@ def create_generators(args):
         train_generator = OpenImagesGenerator(
             args.main_dir,
             subset='train',
+            version=args.version,
             labels_filter=args.labels_filter,
             transform_generator=transform_generator,
             batch_size=args.batch_size
@@ -177,6 +178,7 @@ def create_generators(args):
             validation_generator = OpenImagesGenerator(
                 args.main_dir,
                 subset='validation',
+                version=args.version,
                 labels_filter=args.labels_filter,
                 batch_size=args.batch_size
             )
@@ -224,7 +226,8 @@ def parse_args(args):
 
     oid_parser = subparsers.add_parser('oid')
     oid_parser.add_argument('main_dir', help='Path to dataset directory')
-    oid_parser.add_argument('subset', help='train/validation/test')
+    oid_parser.add_argument('subset', help='Argument for loading a subset from train/validation/test.')
+    oid_parser.add_argument('--version',  help='The current dataset version is V3.', default='2017_11')
     oid_parser.add_argument('--labels_filter',  help='A list of labels to filter.', default=None)
 
     csv_parser = subparsers.add_parser('csv')
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -229,6 +229,8 @@ def parse_args(args):
     oid_parser.add_argument('subset', help='Argument for loading a subset from train/validation/test.')
     oid_parser.add_argument('--version',  help='The current dataset version is V3.', default='2017_11')
     oid_parser.add_argument('--labels_filter',  help='A list of labels to filter.', default=None)
+    oid_parser.add_argument('--annotation_cache_dir', help='Path to store annotation cache.',
+                            default='/media/work2/OpenImages/2017_11')
 
     csv_parser = subparsers.add_parser('csv')
     csv_parser.add_argument('annotations', help='Path to CSV file containing annotations for training.')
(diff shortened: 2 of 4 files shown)


hgaiser:
Ah cool, thanks for the PR!

I'm quite busy with deadlines so I don't know if I'll be able to look at it this week. Perhaps @yhenon can have a look?

yhenon:
Yeah I can do that. Thanks for the nice looking PR @lvaleriu

lvaleriu:
Welcome! For downloading this dataset I followed the instructions from this link: https://github.com/cvdfoundation/open-images-dataset. (This is already specified in the https://github.com/openimages/dataset repository but i prefer to recall it).

As for the metadata, i'm using the v3 version (from 2017_11) of this dataset . This is hard_coded in the generator's constructor so we might add this as a parameter. They will surely publish new versions of the dataset/annotations, so that might be a good idea to allow users to work on a specific version.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,201 @@
...
+        super(OpenImagesGenerator, self).__init__(**kwargs)
+
+    def size(self):
+        return len(self.annotations)
+
+    def num_classes(self):
+        return len(self.id2labels_dict)
+
+    def name_to_label(self, name):
+        raise NotImplementedError()
```
yhenon: this is easy to implement by just reversing the id2labels_dict, maybe we should have it?

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,201 @@
...
+            y1 = float(row['YMin'])
+            y2 = float(row['YMax'])
+
+            x1_abs = int(round(x1 * width))
+            x2_abs = int(round(x2 * width))
+            y1_abs = int(round(y1 * height))
+            y2_abs = int(round(y2 * height))
+
+            if y2_abs <= y1_abs:
+                print ('{} w: {} h: {}'.format(row, width, height))
```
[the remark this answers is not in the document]
lvaleriu: This is not really an error, but more a warning. There are some annotations (2 or 3 i think) in the dataset that are very narrow, so when rounding the X or Y the values become equal so i filter them. Should I put a different warning?

keras_retinanet/preprocessing/open_images.py
lvaleriu: In this dataset some labels (like Bench, for ex) can have different ids since they mean different things. So does it have much sense to implement this method?

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,201 @@
...
+                id_annotations[img_id] = {'w': width, 'h': height, 'boxes': [annotation]}
+    return id_annotations
+
+
+class OpenImagesGenerator(Generator):
+    def __init__(
+            self, main_dir, subset, labels_filter,
+            **kwargs
+    ):
+        metadata_dir = os.path.join(main_dir, '2017_11')
```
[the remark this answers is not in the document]
lvaleriu: Ok!

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,201 @@
...
+        fname_json = os.path.join(metadata_dir, subset, subset + '.json')
```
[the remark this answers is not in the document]
lvaleriu: Ok!

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,201 @@
+import csv
+import json
+import os
+
+import numpy as np
+from PIL import Image
+from tqdm import tqdm
```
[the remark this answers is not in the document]
lvaleriu: Ok!

lvaleriu:
I've pushed a new commit with the different changes.

keras_retinanet/preprocessing/open_images.py
de-vri-es: A quick grep through the repository shows that `name_to_label` is never used outside of generators themselves. With that in mind I think it's fine to keep it unimplemented.

I'm half thinking we should remove it from the base generator API. If there are datasets where there is no 1:1 mapping between class names and ID's, the function could lead to runtime failures if someone actually tries to use it.

lvaleriu: This happens particulary because the labelnames in OID are things like "/m/04hgtk". And each label has a description (ex: car, person, etc...). Some labenames share the same description, but have different meaning.

lvaleriu: If a function in a class is never called outside the class, yes, it might be a good idea to make it internal/private. And remove it from the base generator

keras_retinanet/bin/train.py
```diff
@@ -203,6 +224,12 @@ def parse_args(args):
     pascal_parser = subparsers.add_parser('pascal')
     pascal_parser.add_argument('pascal_path', help='Path to dataset directory (ie. /tmp/VOCdevkit).')
 
+    oid_parser = subparsers.add_parser('oid')
+    oid_parser.add_argument('main_dir', help='Path to dataset directory')
```
hgaiser: Bit nitpicky, but it's missing a `.` at the end of the help.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
+import csv
```
hgaiser: What is the deal with copyright header @de-vri-es ?

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def get_labels(metadata_dir):
+    trainable_classes_path = os.path.join(metadata_dir, 'classes-bbox-trainable.txt')
+    description_path = os.path.join(metadata_dir, 'class-descriptions.csv')
+
+    description_table = {}
+    with open(description_path) as f:
+        for row in csv.reader(f):
+            if len(row):
```
hgaiser: Are there also empty rows? Otherwise I'd remove this.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+                description_table[row[0]] = row[1].replace("\"", "").replace("'", "").replace('`', '')
+
+    with open(trainable_classes_path, 'rb') as f:
+        trainable_classes = f.read().split('\n')
+
+    id2labels_dict = dict([(i, description_table[c]) for i, c in enumerate(trainable_classes)])
```
hgaiser: I prefer the name `id_to_labels` here.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+    cls_index_dict = dict([(c, i) for i, c in enumerate(trainable_classes)])
```
hgaiser: Similarly here, `cls_index`.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+    with open(annotations_path, 'r') as csvfile:
+        reader = csv.DictReader(csvfile,
+                                fieldnames=['ImageID', 'Source', 'LabelName',
+                                            'Confidence', 'XMin', 'XMax', 'YMin',
+                                            'YMax'])
+        reader.next()
+        for _ in reader:
+            cnt += 1
+
+        print ('There are {} lines in subset: {}'.format(cnt, subset))
```
hgaiser: Remove the space between `print` and `(`. I'm wondering if we should print at all in generators.. The COCO generator does so because the underlying library prints, but I prefer no printing unless it's a warning.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+
+    id_annotations = dict()
```
hgaiser: By the way you can also initialize a dict with `id_annotations = {}` which in my opinion is just as clear, but it is up to you what you prefer.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+            frame = row['ImageID']
+
+            class_name = row['LabelName']
+
+            if class_name not in cls_index_dict:
+                continue
+
+            cls_id = cls_index_dict[class_name]
+
+            img_path = os.path.join(main_dir, 'images', subset, frame + '.jpg')
```
hgaiser: I assume all images in the OID are `jpg` ? In case someone wants to use the OID format, but use `png` files, do you think it would be nice to have the extension configurable (but defaulting to `jpg`) ?

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+                        images_sizes[frame] = (width, height)
+                except Exception:
+                    continue
+
+            x1 = float(row['XMin'])
+            x2 = float(row['XMax'])
+            y1 = float(row['YMin'])
+            y2 = float(row['YMax'])
+
+            x1_abs = int(round(x1 * width))
```
hgaiser: Why the name `_abs` ? Short for absolute? They aren't absolute values though.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+            y2_abs = int(round(y2 * height))
+
+            # Check that the bounding box is valid.
+            if x2 <= x1:
+                raise ValueError('line {}: x2 ({}) must be higher than x1 ({})'.format(line, x2, x1))
+            if y2 <= y1:
+                raise ValueError('line {}: y2 ({}) must be higher than y1 ({})'.format(line, y2, y1))
+
+            if y2_abs == y1_abs:
+                print ('filtering line {}: rounding y2 ({}) and y1 ({}) makes them equal'.format(line, y2, y1))
```
hgaiser: I prefer to print warnings with `warnings`, like [here](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/preprocessing/generator.py#L101).

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+            if x2_abs == x1_abs:
+                print ('filtering line {}: rounding x2 ({}) and x1 ({}) makes them equal'.format(line, x2, x1))
+                continue
+
+            img_id = row['ImageID']
+
+            annotation = {'cls_id': cls_id, 'x1': x1, 'x2': x2, 'y1': y1, 'y2': y2}
+
+            if img_id in id_annotations:
+
```
hgaiser: Empty line.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+                annotations = id_annotations[img_id]
+                annotations['boxes'].append(annotation)
+            else:
+                id_annotations[img_id] = {'w': width, 'h': height, 'boxes': [annotation]}
+    return id_annotations
+
+
+class OpenImagesGenerator(Generator):
+    def __init__(
+            self, main_dir, subset, version, labels_filter,
```
hgaiser: `version` could have the same default here as in `train.py`. As far as I understood it, that default makes sense?

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+            **kwargs
+    ):
+        self.base_dir = os.path.join(main_dir, 'images', subset)
+
+        metadata_dir = os.path.join(main_dir, version)
+        annotation_cache_json = os.path.join(metadata_dir, subset, subset + '.json')
+        labels_dict = None if labels_filter is None else dict([(l, i) for i, l in enumerate(labels_filter)])
+
+        print ('loading {} subset'.format(subset))
```
hgaiser: Preferably no printing.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,206 @@
...
+        self.id2labels_dict, cls_index_dict = get_labels(metadata_dir)
+
+        if os.path.exists(annotation_cache_json):
+            with open(annotation_cache_json, 'r') as f:
+                self.annotations = json.loads(f.read())
+        else:
+            self.annotations = generate_images_annotations_json(main_dir, metadata_dir, subset, cls_index_dict)
+            json.dump(self.annotations, open(annotation_cache_json, "w"))
```
hgaiser: It might not be possible to write in this location, should we catch this and handle it appropriately?

keras_retinanet/preprocessing/open_images.py
lvaleriu: you're right!

keras_retinanet/preprocessing/open_images.py
lvaleriu: ok!

keras_retinanet/preprocessing/open_images.py
lvaleriu: ok!

keras_retinanet/bin/train.py
lvaleriu: Np :)

keras_retinanet/preprocessing/open_images.py
lvaleriu: The last line in the class-descriptions csv is empty

lvaleriu: i'll add some commentary in the code

keras_retinanet/preprocessing/open_images.py
lvaleriu: ok

keras_retinanet/preprocessing/open_images.py
lvaleriu: I added a default value for version, corresponding to the V3 of the dataset. And for labels_filter too, since by default we want to train on the whole dataset.

keras_retinanet/preprocessing/open_images.py
lvaleriu: ok

keras_retinanet/preprocessing/open_images.py
lvaleriu: Yes. They are all JPGs. 
It would be better to have a list of extension to scan eventually, but i prefer to have more feedback before doing that.

keras_retinanet/preprocessing/open_images.py
lvaleriu: I'll add a new parameter with the default value equal to '.' i.e. current directory

keras_retinanet/preprocessing/open_images.py
de-vri-es: There should be a copyright notice and license indication as with the other files in the repository.

@lvaleriu: Could you add them, and make sure to attribute copyright to the proper peson/company. That is most likely you, unless you made the changes during work hours for your employer (then it should be your employer).

lvaleriu: Done!

keras_retinanet/preprocessing/open_images.py
```diff
@@ -0,0 +1,232 @@
...
+            y2_int = int(round(y2 * height))
+
+            # Check that the bounding box is valid.
+            if x2 <= x1:
+                raise ValueError('line {}: x2 ({}) must be higher than x1 ({})'.format(line, x2, x1))
+            if y2 <= y1:
+                raise ValueError('line {}: y2 ({}) must be higher than y1 ({})'.format(line, y2, y1))
+
+            if y2_int == y1_int:
+                print ('filtering line {}: rounding y2 ({}) and y1 ({}) makes them equal'.format(line, y2, y1))
```
hgaiser: Use `warnings.warn` here too.

hgaiser:
If you address the last remark and add an explanation to the README, I think it is ready to be merged :)

keras_retinanet/preprocessing/open_images.py
lvaleriu: Yep. Wondering how i missed that! :)

lvaleriu:
That should be ok!

README.md
```diff
@@ -59,6 +59,19 @@ The pretrained MS COCO model can be downloaded [here](https://github.com/fizyr/k
...
 ```
 
+For training on [OID](https://github.com/openimages/dataset), run:
+```shell
+# Running directly from the repository:
+keras_retinanet/bin/train.py oid /path/to/OID 2017_11
+
+# Using the installed script:
+retinanet-train oid /path/to/OID 2017_11
+
```
hgaiser: Shouldn't the shell part end here?

README.md
```diff
@@ -59,6 +59,19 @@ The pretrained MS COCO model can be downloaded [here](https://github.com/fizyr/k
...
+You can specify a list of labels if you want to train on a subset
+by adding the parameter labels_filter:
+    labels_filter = ["Helmet", "Tree", "Stop sign", "Traffic light"]
```
hgaiser: Add this where/how ?

lvaleriu: In the shell script arguments.

lvaleriu: I'm updating the readme

lvaleriu:
Modified the readme & added debug option for the OID generator too (along  the others ... )

hgaiser:
Thanks!

PR #255, opened 2018-01-30
No use of model.output[3:]?

jaydu1:
model.outputs[3:] = []


hgaiser:
Hi, thanks for the PR. It's true, with the default settings `model.outputs[3:]` is empty. However the design of `keras-retinanet` is such that you can extend it with custom submodels. This way you can train other things than class label and bbox regression (we've used it to compute a rough orientation in the past). If you have 3 or more submodels, `models.outputs,[3:]` is no longer empty. If you have 2 submodels it is empty, but then it just doesn't do anything so there is no harm.

PR #262, opened 2018-01-31
Add visualization of anchors and anchored annotations in debug.py.

hgaiser:
This PR adds the ability to visualize anchors and something I named anchored annotation (lack of better term?). The latter are annotations which have anchors "pointing" to them (see the example below).

screenshot from 2018-01-31 20-21-06

In that image, you notice some annotations have a yellow edge around it. Those annotations have at least one anchor that is trying to regress to that annotation. The other annotations that have only a green edge have no anchor attached to it (most likely they are too small) and therefore don't contribute to the training process (even though one might expect them to be contributing when looking purely at the annotations).

This can be useful for debugging purposes.

### DIFF
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -20,6 +20,7 @@ import argparse
 import os
 import sys
 import cv2
+import numpy as np
 
 # Allow relative imports when being executed as script.
 if __name__ == "__main__" and __package__ is None:
@@ -31,7 +32,7 @@ if __name__ == "__main__" and __package__ is None:
 from ..preprocessing.pascal_voc import PascalVocGenerator
 from ..preprocessing.csv_generator import CSVGenerator
 from ..utils.transform import random_transform_generator
-from ..utils.visualization import draw_annotations
+from ..utils.visualization import draw_annotations, draw_boxes
 
 
 def create_generator(args):
@@ -94,7 +95,8 @@ def parse_args(args):
     csv_parser.add_argument('classes',     help='Path to a CSV file containing class label mapping.')
 
     parser.add_argument('--no-resize', help='Disable image resizing.', dest='resize', action='store_false')
-    parser.add_argument('--annotations', help='Show annotations on the image.', action='store_true')
+    parser.add_argument('--anchors', help='Show positive anchors on the image.', action='store_true')
+    parser.add_argument('--annotations', help='Show annotations on the image. Green annotations have anchors, red annotations don\'t and therefore don\'t contribute to training.', action='store_true')
     parser.add_argument('--random-transform', help='Randomly transform image and annotations.', action='store_true')
 
     return parser.parse_args(args)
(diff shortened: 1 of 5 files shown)


awilliamson:
Fantastic! So you're saying the yellow based anchors have this 'field' around them, where those points pull towards that regression target?
Might it be visually more beneficial to label the examples which do not contribute in red, or some other colour to distinguish them apart?

I will definitely have to have a look at this :)

hgaiser:
> Fantastic! So you're saying the yellow based anchors have this 'field' around them, where those points pull towards that regression target?

Yes, the `--anchors` option shows that, but I felt it was a bit unclear which annotations are "anchored" and which ones weren't.

screenshot from 2018-01-31 20-31-02

> Might it be visually more beneficial to label the examples which do not contribute in red, or some other colour to distinguish them apart?

Would be nice maybe, but simple as it may sound it requires a bit of a different approach (most simply by looking up the annotations in the list of regressed anchors and drawing the ones that don't have a match). The current method simply draw all regressed anchors, which is a bit of an overkill maybe but it's a debug script anyway, and thereby it "labels" the annotations.

de-vri-es:
> Would be nice maybe, but simple as it may sound it requires a bit of a different approach (most simply by looking up the annotations in the list of regressed anchors and drawing the ones that don't have a match). The current method simply draw all regressed anchors, which is a bit of an overkill maybe but it's a debug script anyway, and thereby it "labels" the annotations.

Or we could draw all annotations in red, and then draw the regressed anchors in green over the annotations (they seem to be drawn behind the annotations now). A bit hackish maybe, but should work =]

hgaiser:
> Or we could draw all annotations in red, and then draw the regressed anchors in green over the annotations (they seem to be drawn behind the annotations now). A bit hackish maybe, but should work =]

Hehe yeah true, that'd work.

hgaiser:
If I draw the regressed anchors (renamed anchored annotations to regressed anchors, it's more accurate) after the annotations, I get the following image:

screenshot from 2018-01-31 20-41-45

Would this make sense? Or are the colors confusing?

de-vri-es:
I think it's a bit confusing to have green be bad and yellow be good. I'd go for red for annotations and green for regressed anchors. I don't think it's a problem to have red annotations when the regressed anchors aren't shown.

Or we could just always draw regressed anchors. Performance for the debug script isn't really a big issue here I think.

hgaiser:
> I think it's a bit confusing to have green be bad and yellow be good.

yeah, I agree

> Or we could just always draw regressed anchors. Performance for the debug script isn't really a big issue here I think.

As in always draw the regressed anchors when annotations are drawn? In other words, removing the option to display them separately.

de-vri-es:
> As in always draw the regressed anchors when annotations are drawn? In other words, removing the option to display them separately?

Yeah

hgaiser:
Latest version: 
screenshot from 2018-01-31 20-51-38

I removed the option, annotations are now drawn in green if they have anchors, red if they don't.

PR #263, opened 2018-02-01
Snapshot path existence check

awilliamson:
Creates the given snapshot path if it doesn't exist.

h5py would error out at the end of an epoch if the snapshot directory doesn't exist. Behaviour for things like Tensorboard logging are to create directories if they don't exist. As these are explicit arguments, it makes sense to have that default creation behaviour.

Let me know what you think?


hgaiser:
Cool didn't know about `exist_ok`. LGTM!

de-vri-es:
Small but very useful addition. Thanks!

PR #275, opened 2018-02-07
Fix training without weights.

de-vri-es:
Currently, `train.py` doesn't allow training without weights because the command line arguments are passed down directly, and the default is "imagenet". That means there is no way to *not* load any weights right now.

This PR does two things:
* It moves loading of weights out of the retinanet-resnet model creation functions.
* It removes all special string values for weights loading. Instead, different flags indicate different types of weight initialization: `--snapshot <file>`, `--imagenet-weights`, `--weights <file>` or `--no-weights`. The default is still to initialize with imagenet weights.

Note: I only tested the argument parsing briefly. When I get back to the office (next week) I intend to test this more thoroughly to verify the actual runtime behaviour. But it can be reviewed already.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -250,8 +250,9 @@ def parse_args(args):
 
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--snapshot',          help='Resume training from a snapshot.')
-    group.add_argument('--imagenet-weights',  help='Initialize the model with pretrained imagenet weights. This is the default behaviour.', action='store_true', default=True)
+    group.add_argument('--imagenet-weights',  help='Initialize the model with pretrained imagenet weights. This is the default behaviour.', action='store_const', const=True, default=True)
     group.add_argument('--weights',           help='Initialize the model with weights from a file.')
+    group.add_argument('--no-weights',        help='Don\'t initialize the model with any weights.', dest='imagenet_weights', action='store_const', const=False)
 
     parser.add_argument('--batch-size',    help='Size of the batches.', default=1, type=int)
     parser.add_argument('--gpu',           help='Id of the GPU to use (as reported by nvidia-smi).')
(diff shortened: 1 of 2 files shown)


keras_retinanet/bin/train.py
```diff
@@ -39,25 +39,34 @@
 from ..preprocessing.pascal_voc import PascalVocGenerator
 from ..preprocessing.csv_generator import CSVGenerator
 from ..preprocessing.open_images import OpenImagesGenerator
-from ..models.resnet import resnet50_retinanet, custom_objects
+from ..models.resnet import resnet_retinanet, custom_objects, download_imagenet
 from ..utils.transform import random_transform_generator
 from ..utils.keras_version import check_keras_version
 
 
+RESNET_BACKBONE = 50
```
hgaiser: Maybe we should make this an argument?

keras_retinanet/bin/train.py
```diff
@@ -273,14 +284,21 @@ def main(args=None):
...
 
     # create the model
-    if args.snapshot:
+    if args.snapshot is not None:
         print('Loading model, this may take a second...')
         model            = keras.models.load_model(args.snapshot, custom_objects=custom_objects)
         training_model   = model
         prediction_model = model
     else:
+        weights = None
```
hgaiser: Isn't this the same as:
```python
weights = args.weights
if args.imagenet_weights:
    weights = download_imagenet(RESNET_BACKBONE)
```
?

keras_retinanet/bin/train.py
de-vri-es: We could do that. I'd prefer to do that in a separate PR though, and keep this one focussed on weights loading.

keras_retinanet/bin/train.py
de-vri-es: Hmm yeah, it is if you know that --weights and --imagenet-weights are mutually exclusive. I suppose that is a reasonable thing to assume here.

de-vri-es: Actually, it isn't. `args.imagenet_weights` is always true, except when `--no-weights` is specified. But it can be simplified to:
```
weights = args.weights
if weights is None and args.imagenet_weights:
    weights = download_imagenet(RESNET_BACKBONE)
```

keras_retinanet/bin/train.py
hgaiser: Alright, then this LGTM :)

PR #281, opened 2018-02-10
Refactor debug.py to more closely use generator output.

hgaiser:
This PR refactors `debug.py` such that it uses the output of the generator as closely as currently possible. This also means that the debug script can now visualize `batch_size > 1`, which is what I intended it for in the first place.

### DIFF
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -234,7 +234,7 @@ class Generator(object):
         # compute network targets
         targets = self.compute_targets(image_group, annotations_group)
 
-        return inputs, targets
+        return inputs, targets, annotations_group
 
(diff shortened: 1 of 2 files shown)


keras_retinanet/preprocessing/generator.py
```diff
@@ -234,7 +234,7 @@ def compute_input_output(self, group):
         # compute network targets
         targets = self.compute_targets(image_group, annotations_group)
 
-        return inputs, targets
+        return inputs, targets, annotations_group
```
de-vri-es: I think we should be careful with the generator API. Returning the annotations here seems like a bit of a hack. I'd like to discuss this at the office a bit.

hgaiser:
Closing this since it's so old. Might look at it again at some point.

PR #286, opened 2018-02-13
Implement new backbone => MobileNet

lvaleriu:


keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+    model_name = 'mobilenet_%s_%d_tf_no_top.h5' % (alpha_text, rows)
+    weigh_path = BASE_WEIGHT_PATH + model_name
+    weights_path = get_file(model_name,
+                            weigh_path,
+                            cache_subdir='models')
+
+    return weights_path
+
+
+def mobilenet_retinanet(num_classes, backbone=224, inputs=None, **kwargs):
```
hgaiser: `backbone` is not used?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+    alpha = 1.0
```
hgaiser: Should this be configurable? I haven't read much on MobileNet yet but from what I remember this defines the complexity of the model?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+        inputs = keras.layers.Input((None, None, 3))
+
+    mobilenet = MobileNet(input_tensor=inputs, alpha=alpha, include_top=False, pooling=None, weights=None)
+    outputs = [mobilenet.get_layer(name='conv_pw_{}_relu'.format(i)).output for i in [3, 5, 11, 13]]
+
+    # create the mobilenet backbone
+    mobilenet = Model(inputs=inputs, outputs=outputs, name=mobilenet.name)
+
+    # create the full model
+    model = retinanet.retinanet_bbox(inputs=inputs,
```
hgaiser: I know it's not default PEP8, but I much more prefer the alternative style for multi-line function calling (which is also permitted by PEP8):
```python
foo(
    bar,
    baz=True
)
```

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+    elif alpha == 0.75:
+        alpha_text = '7_5'
+    elif alpha == 0.50:
+        alpha_text = '5_0'
+    else:
+        alpha_text = '2_5'
+
+    model_name = 'mobilenet_%s_%d_tf_no_top.h5' % (alpha_text, rows)
+    weigh_path = BASE_WEIGHT_PATH + model_name
+    weights_path = get_file(model_name,
```
hgaiser: See below for multi-line function calling.

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+    if alpha == 1.0:
+        alpha_text = '1_0'
+    elif alpha == 0.75:
+        alpha_text = '7_5'
+    elif alpha == 0.50:
+        alpha_text = '5_0'
+    else:
+        alpha_text = '2_5'
+
+    model_name = 'mobilenet_%s_%d_tf_no_top.h5' % (alpha_text, rows)
```
hgaiser: Prefer to use `.format()` for formatting.

https://stackoverflow.com/questions/5082452/python-string-formatting-vs-format

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+    weigh_path = BASE_WEIGHT_PATH + model_name
```
hgaiser: `weigh_path` typo?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+from keras.models import Model
+from keras.applications.mobilenet import MobileNet, BASE_WEIGHT_PATH, get_file
+
+from ..models import retinanet
+
+custom_objects = retinanet.custom_objects.copy()
+
+
+def download_imagenet(backbone):
+    alpha = 1.0
```
hgaiser: Make `alpha` configurable? Can it only be in `[1.0, 0.75, 0.50, 0.25]`? If so, we should maybe check for that?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+    rows = backbone
```
hgaiser: What is `backbone`/`rows`?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
+import keras
+from keras import backend as K
+from keras.models import Model
+from keras.applications.mobilenet import MobileNet, BASE_WEIGHT_PATH, get_file
+
+from ..models import retinanet
+
+custom_objects = retinanet.custom_objects.copy()
```
hgaiser: No need for a copy.

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def download_imagenet(backbone):
+    alpha = 1.0
+    rows = backbone
+
+    # load weights
+    if K.image_data_format() == 'channels_first':
+        raise ValueError('Weights for "channels_last" format '
```
hgaiser: I'm fine with this being one line.

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
+import keras
+from keras import backend as K
```
hgaiser: Considering we use `keras.backend` everywhere, I prefer that here too. Actually that means this import can be removed.

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,56 @@
...
+from keras.models import Model
```
hgaiser: So can this one as far as I'm concerned.

keras_retinanet/bin/train.py
```diff
@@ -315,10 +324,14 @@ def main(args=None):
         weights = args.weights
         # default to imagenet if nothing else is specified
         if weights is None and args.imagenet_weights:
-            weights = download_imagenet(RESNET_BACKBONE)
+            weights = download_imagenet(args.backbone_param)
 
         print('Creating model, this may take a second...')
-        model, training_model, prediction_model = create_models(num_classes=train_generator.num_classes(), weights=weights, multi_gpu=args.multi_gpu)
+        model, training_model, prediction_model = create_models(backbone_retinanet=retinanet,
```
hgaiser: Same thing regarding multi-line function calls (see below).

keras_retinanet/bin/train.py
```diff
@@ -305,6 +307,13 @@ def main(args=None):
     # create the generators
     train_generator, validation_generator = create_generators(args)
 
+    if args.backbone == 'resnet':
+        from ..models.resnet import resnet_retinanet as retinanet, custom_objects, download_imagenet
+    elif args.backbone == 'mobilenet':
+        from ..models.mobilenet import mobilenet_retinanet as retinanet, custom_objects, download_imagenet
+    else:
+        raise NotImplementedError('Backbone model not implemented.')
```
hgaiser: Also print the `backbone` here, maybe something like:
```python
raise NotImplementedError('Backbone \'{}\' not implemented.'.format(args.backbone))
```

keras_retinanet/bin/train.py
```diff
@@ -275,6 +275,8 @@ def csv_list(string):
     group.add_argument('--weights',           help='Initialize the model with weights from a file.')
     group.add_argument('--no-weights',        help='Don\'t initialize the model with any weights.', dest='imagenet_weights', action='store_const', const=False)
 
+    parser.add_argument('--backbone',        help='Backbone model used by retinanet.', default='resnet', type=str)
+    parser.add_argument('--backbone_param',  help='Backbone model parameter.', default='50', type=int)
```
hgaiser: Prefer to use `-` instead of `_` (so `--backbone-param`).

Also, @de-vri-es , can you give your opinion on this part? I know you have some thoughts about it.

keras_retinanet/bin/train.py
```diff
@@ -59,14 +55,15 @@ def model_with_weights(model, weights, skip_mismatch):
...
-def create_models(num_classes, weights, multi_gpu=0):
+def create_models(backbone_retinanet, backbone_param, num_classes, weights, multi_gpu=0):
     # create "base" model (no NMS)
 
     # Keras recommends initialising a multi-gpu model on the CPU to ease weight sharing, and to prevent OOM errors.
     # optionally wrap in a parallel model
     if multi_gpu > 1:
         with tf.device('/cpu:0'):
-            model = model_with_weights(resnet_retinanet(num_classes, backbone=RESNET_BACKBONE, nms=False), weights=weights, skip_mismatch=True)
+            model = model_with_weights(backbone_retinanet(num_classes, backbone=backbone_param, nms=False),
```
hgaiser: Multi-line function calls (see below).

keras_retinanet/bin/train.py
```diff
@@ -76,7 +73,8 @@ def create_models(num_classes, weights, multi_gpu=0):
         detections       = layers.NonMaximumSuppression(name='nms')([boxes, classification, detections])
         prediction_model = keras.models.Model(inputs=model.inputs, outputs=model.outputs[:2] + [detections])
     else:
-        model            = model_with_weights(resnet_retinanet(num_classes, backbone=RESNET_BACKBONE, nms=True), weights=weights, skip_mismatch=True)
+        model            = model_with_weights(backbone_retinanet(num_classes, backbone=backbone_param, nms=True),
```
hgaiser: Multi-line function calls (see below).

awilliamson:
Quite a lot of README updates there, that aren't overly descriptive or distinct. Perhaps squash a few of those? And the two additional 'merge' commits which shouldn't be needed.

keras_retinanet/preprocessing/open_images.py
```diff
@@ -1,3 +1,19 @@
+"""
```
awilliamson: How is this classed as an addition from the commit? As the current master has this license in place for this exact file?

keras_retinanet/models/mobilenet.py
lvaleriu: You're right!

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,62 @@
...
+        alpha_text = '7_5'
+    elif alpha == 0.50:
+        alpha_text = '5_0'
+    else:
+        alpha_text = '2_5'
+
+    model_name = 'mobilenet_%s_%d_tf_no_top.h5' % (alpha_text, rows)
+    weigh_path = BASE_WEIGHT_PATH + model_name
+    weights_path = get_file(model_name,
+                            weigh_path,
```
awilliamson: `weigh` here too.

keras_retinanet/bin/train.py
awilliamson: Personally, I think this should be a singular parameter. We refer to models by ResNet50, ResNet101 etc. And the `param` number may not hold for different backbones. It's far better to use singular names for things, rather than guessing that all conceivable models are `name + depth` based.

keras_retinanet/bin/train.py
```diff
@@ -305,6 +307,13 @@ def main(args=None):
     # create the generators
     train_generator, validation_generator = create_generators(args)
 
+    if args.backbone == 'resnet':
+        from ..models.resnet import resnet_retinanet as retinanet, custom_objects, download_imagenet
```
awilliamson: Where does custom_objects come from? The import was removed?

awilliamson: Likewise for download_imagenet.

keras_retinanet/bin/train.py
de-vri-es: I don't really like splitting the backbone definition into two parameters. For example: I'd prefer `--backbone resnet50` over `--backbone resnet --backbone-param 50`. 

That also helps catching errors in the command line options quickly: `resnet42` shouldn't be accepted. `parse_args` and/or `check_args` should fail on that.

keras_retinanet/models/mobilenet.py
lvaleriu: The parameters of this network (when loading imagenet) are:
        alpha: controls the width of the network.
            - If `alpha` < 1.0, proportionally decreases the number
                of filters in each layer.
            - If `alpha` > 1.0, proportionally increases the number
                of filters in each layer.
            - If `alpha` = 1, default number of filters from the paper
                 are used at each layer.
            - If imagenet weights are being loaded, alpha can be one of' `0.25`, `0.50`, `0.75` or `1.0` only.'
      depth_multiplier 
          -if imagenet weights are being loaded, 'depth multiplier must be 1.
      input_shape: If imagenet weights are being loaded, '
                             'input must have a static square shape (one of '
                             '(128,128), (160,160), (192,192), or (224, 224))

So they are different from the resnet.

keras_retinanet/bin/train.py
awilliamson: Wrap these imports in try and excepts, maybe? As it could fail to import, and that should count as 'NotImplemented' too?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,62 @@
+import keras
```
de-vri-es: Thanks for the PR!

Could you add a license and copyright notice to the top of the file?

keras_retinanet/models/mobilenet.py
lvaleriu: As they are different from the resnet we should refactor the backbone loading in the train.py script. I already started, but I prefer to let you finish this. This issue will appear again when i'll do the next pull request with the Densenet backbone.

keras_retinanet/models/mobilenet.py
lvaleriu: Yes, they should be configurable. But before doing that we should refactor the backbone initialisation and consider that backbones could have different parameters.

awilliamson:
@lvaleriu When making changes to existing commits, it might be worth doing these as amended commits, rather than fresh commits where possible. Especially for non-functional related changes. This just makes things cleaner and easier to understand.

keras_retinanet/models/mobilenet.py
de-vri-es: Yeah, we should probably have some better way to configure a backbone including more detailed configuration parameters. We might have to rework command line parsing to really support that nicely.

keras_retinanet/models/mobilenet.py
lvaleriu: Didnt really understand this. Can we have: 
foo(
    bar, bar2,
    baz=True,
    baz2=False
) ?

keras_retinanet/bin/train.py
lvaleriu: done!

lvaleriu:
You are right. I'll do that for the next commits and try to fix the existing.

keras_retinanet/bin/train.py
lvaleriu: You are right. I'm fixing this and in the same time doing a small refactoring. Eager to hear your opinion.

keras_retinanet/models/mobilenet.py
lvaleriu: done

keras_retinanet/models/mobilenet.py
lvaleriu: Did some work on this in the latest commit.

lvaleriu:
With some help from a friend i squashed and rebased my commits for this pull request.

keras_retinanet/models/mobilenet.py
hgaiser: As far as I'm concerned, yeah. Not sure what PEP8 says though.

lvaleriu:
Added some doc in the readme. @hgaiser

README.md
```diff
@@ -23,6 +23,12 @@ If you installed `keras-retinanet` correctly, the train script will be installed
 However, if you make local modifications to the `keras-retinanet` repository, you should run the script directly from the repository.
 That will ensure that your local changes will be used by the train script.
 
+The default backbone is 'resnet50'. You can change this using the '--backbone=xxx' argument in the running script where
+xxx can be one of the backbones in resnet models (resnet50, resnet101, resnet152) or 
+mobilenet models (mobilenet128_1.0, mobilenet128_0.75, mobilenet160_1.0, etc...).
+The different options are defined by each model in their corresponding python scripts (resnet.py, mobilenet.py) 
+More backbones will be added.
```
hgaiser: I think it's better to remove this line.

keras_retinanet/bin/train.py
```diff
@@ -315,10 +328,14 @@ def main(args=None):
         weights = args.weights
         # default to imagenet if nothing else is specified
         if weights is None and args.imagenet_weights:
-            weights = download_imagenet(RESNET_BACKBONE)
+            weights = download_imagenet(args.backbone)
 
         print('Creating model, this may take a second...')
-        model, training_model, prediction_model = create_models(num_classes=train_generator.num_classes(), weights=weights, multi_gpu=args.multi_gpu)
+        model, training_model, prediction_model = create_models(backbone_retinanet=retinanet,
```
hgaiser: Reshape as:
```python
foo(
    bar,
    baz
)
```

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,82 @@
...
+        alpha_text = '1_0'
+    elif alpha == 0.75:
+        alpha_text = '7_5'
+    elif alpha == 0.50:
+        alpha_text = '5_0'
+    else:
+        alpha_text = '2_5'
+
+    model_name = 'mobilenet_{}_{}_tf_no_top.h5'.format(alpha_text, rows)
+    weight_path = BASE_WEIGHT_PATH + model_name
```
hgaiser: What's the difference between `weight_path` and `weights_path` ?

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,82 @@
...
+        inputs = keras.layers.Input((None, None, 3))
+
+    mobilenet = MobileNet(input_tensor=inputs, alpha=alpha, include_top=False, pooling=None, weights=None)
+    outputs = [mobilenet.get_layer(name='conv_pw_{}_relu'.format(i)).output for i in [3, 5, 11, 13]]
+
+    # create the mobilenet backbone
+    mobilenet = keras.models.Model(inputs=inputs, outputs=outputs, name=mobilenet.name)
+
+    # create the full model
+    model = retinanet.retinanet_bbox(inputs=inputs,
```
hgaiser: Reformat as:
```python
foo(
    bar,
    baz
)
```

README.md
lvaleriu: And I was so happy to add this :). Ok, np :)

hgaiser: Hehe I only meant the last line though.

keras_retinanet/models/mobilenet.py
lvaleriu: No difference - just a copy/paste. I'm removing it.

lvaleriu: "weight_path" is in fact "weights_url". Changing this!

keras_retinanet/bin/train.py
lvaleriu: model, training_model, prediction_model = create_models(backbone_retinanet=retinanet,
                                                                backbone=args.backbone,
                                                                num_classes=train_generator.num_classes(),
                                                                weights=weights,
                                                                multi_gpu=args.multi_gpu)  
=>
      model, training_model, prediction_model = create_models(backbone_retinanet=retinanet, backbone=args.backbone, num_classes=train_generator.num_classes(), weights=weights, multi_gpu=args.multi_gpu)

?

hgaiser:
```python
      model, training_model, prediction_model = create_models(backbone_retinanet=retinanet,
                                                                backbone=args.backbone,
                                                                num_classes=train_generator.num_classes(),
                                                                weights=weights,
                                                                multi_gpu=args.multi_gpu)  
```
to:
```python
model, training_model, prediction_model = create_models(
    backbone_retinanet=retinanet,
    backbone=args.backbone,
    num_classes=train_generator.num_classes(),
    weights=weights,
    multi_gpu=args.multi_gpu
)
```

README.md
```diff
@@ -23,6 +23,8 @@ If you installed `keras-retinanet` correctly, the train script will be installed
 However, if you make local modifications to the `keras-retinanet` repository, you should run the script directly from the repository.
 That will ensure that your local changes will be used by the train script.
 
+The default backbone is 'resnet50'. You can change this using the '--backbone=xxx' argument in the running script.
```
hgaiser: Ah apologies, I actually only meant the part saying "More backbones will be added in the future". The list of accepted backbones is useful here I'd say.

hgaiser:
Thanks for your additions. Could you add documentation to the added functions? I know not every function has it at the moment, but I'm trying to make a habit out of it :).

Also, please add a unit test that tests mobilenet: https://github.com/fizyr/keras-retinanet/blob/master/tests/bin/test_train.py

Once that's done, this can be merged as far as I'm concerned. @awilliamson @de-vri-es anything to add?

README.md
lvaleriu: Ok!

lvaleriu:
Do you have an example for unit testing the training of retinanet? What do you expect exactly? Unit testing of a specific backbone or the ensemble - resinet with backbone?

You are also talking about "functions"? What functions exactly? I have some comments in the code for the important steps already.

lvaleriu:
Should i modify the existing test_train script and enumerate all the existing backbones and run the different unit tests on each backbone?

hgaiser:
> Do you have an example for unit testing the training of retinanet? What do you expect exactly? Unit testing of a specific backbone or the ensemble - resinet with backbone?

I would extend the test file [here](https://github.com/fizyr/keras-retinanet/blob/master/tests/bin/test_train.py) to include the different mobilenet architectures. I see that only resnet50 is tested there at the moment, if you want you can also add tests for resnet152 / resnet101.

> You are also talking about "functions"? What functions exactly? I have some comments in the code for the important steps already.

I meant documentation such as is done [here](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/utils/visualization.py).

> Should i modify the existing test_train script and enumerate all the existing backbones and run the different unit tests on each backbone? This should be done in any cases i think.

Yeah, exactly.

lvaleriu:
Done!

hgaiser:
The tests seem to fail. Did they succeed locally? Judging by the logs it seems to be killed, possibly because it ran OOM.

lvaleriu:
Yes, they succeded locally. What I've changed is that i'm testing much more backbones - every resnet, every mobilenet - (not only one) on each test dataset (pascal,  coco, csv).

de-vri-es:
Exit code 137 indicates it was killed by a SIGKILL. The source of the signal could be anything, but I'd hope that if it was travis itself imposing some limits, it would tell us.

It could be memory. If for some reason all models stay loaded while the tests run, that might be a cause.

lvaleriu:
image

Here's the memory profile I have during train tests.

lvaleriu:
image

The "test_pascal" method takes 3m16s now, with all the backends.

lvaleriu:
image

for all the 3 methods ...

de-vri-es:
Why would adding `sudo: required` help?

hgaiser:
> Why would adding sudo: required help?

According to the page I linked to:

> Any of the following sudo or docker related settings route your build to a sudo-enabled linux (Ubuntu Trusty) environment on Google Cloud Engine.
`services: docker`
`sudo: required` or `sudo: true`
any other sudo command in your build script _ any other docker command in your build script

de-vri-es:
Alright, lets give that a try then.

hgaiser:
@lvaleriu could you add `sudo: required` in `.travis.yml` ?

lvaleriu:
I'll try to use @pytest.mark.parametrize("backend", backends) in order to split each backend test in a different method. This should reduce the total time.

lvaleriu:
Any idea?

hgaiser:
LGTM, thanks a lot for your efforts!

hgaiser:
Oh I see you committed everything under "Test only one alpha value". Could you change the commit message to something more meaningful? Perhaps "Implement MobileNet backbone."?

In general having multiple commits is preferable, if they make sense (for instance the unit tests can be a separate commit).

Let me know when you're finished and I'll merge.

lvaleriu:
I've changed the last commit message to "Implement MobileNet backbone.". Yes, you are right, for the next  contributions i'll push more commits (but not too many, since it is difficult to follow).

So it is done.

keras_retinanet/models/mobilenet.py
```diff
@@ -0,0 +1,95 @@
...
+    alpha = float(backbone.split('_')[1])
+
+    # choose default input
+    if inputs is None:
+        inputs = keras.layers.Input((None, None, 3))
+
+    mobilenet = MobileNet(input_tensor=inputs, alpha=alpha, include_top=False, pooling=None, weights=None)
+
+    # get last layer from each depthwise convolution blocks 3, 5, 11 and 13
+    outputs = [mobilenet.get_layer(name='conv_pw_{}_relu'.format(i)).output for i in [3, 5, 11, 13]]
```
kriskorrel-cw, line 87: @lvaleriu 
A little late to the party, but may I ask on what basis these blocks were chosen?
In the issues on this repo I see that many have tried using MobileNet as backbone without great success, and I was wondering what the effect of this choice could be.

The current version of this repo uses `conv_pw_5_relu`, `conv_pw_11_relu` and `conv_pw_13_relu` for `C3`, `C4`, `C5` in the feature pyramid, while for Resnet50, stages 3, 4 and 5 are used. Based on https://github.com/google-research/receptive_field/blob/master/receptive_field/RECEPTIVE_FIELD_TABLE.md there should be quite a difference in receptive field between these choices, visualized in the following table

|  **Feature pyramid level** |  C3 | C4  |  C5 |  
|---|---|---|---|
| **anchor base size** | 32 | 64 | 128 |   
|  **resnet_50_v1 RF** |  99 | 291  |  483 |   
| **mobilenet_v1 TF**  |  43 | 219  |  315 |   

The difference is most prominent for level 3 in the feature pyramid where the receptive field for Resnet50 is more than 3 times the base anchor size (in a single dimension), while the receptive field for MobileNet would only be 34% larger than the base anchor size, meaning that anchors could quite easily become larger than the receptive field of `C3`. I was wondering whether this hinders especially small-scale object detection when using MobileNet as backbone.

There are two important things to note here:
- The effective receptive field (ERF) is often quite smaller even than the theoretical receptive field.
- The top-down and lateral connections in the feature pyramid in theory greatly increase the receptive field for `C3` and `C4`. I assume though that the "original" receptive field without these connections is also quite important.

kriskorrel-cw: Ah I think I see now why these layers were chosen. For future reference: these are the deepest layers for which the effective strides match the default strides for (resnet50-)retinanet.

PR #267, opened 2018-02-02
Added Tensorboard Logging callback

awilliamson:
Additionally added log-dir argparse.

Not sure if this is particularly wanted. But I know quite a few users of Retinanet have been falling back to some Tensorboard logging. This is rather simplistic, but can be expanded upon later for more complex behaviours.

PR #287, opened 2018-02-13
Add mAP to tensorboard when using train.py

mxvs:
If the user enables Tensorboard in `train.py` with `--tensorboard-dir` (introduced with #267) and has not disabled evaluation with `--no-evaluation` the mean_ap value will now be automatically added to TensorBoard (as introduced in #278).

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -109,8 +109,10 @@ def create_callbacks(model, training_model, prediction_model, validation_generat
         checkpoint = RedirectModel(checkpoint, prediction_model)
         callbacks.append(checkpoint)
 
+    tensorboard_callback = None
+
     if args.tensorboard_dir:
-        tensorb = keras.callbacks.TensorBoard(
+        tensorboard_callback = keras.callbacks.TensorBoard(
             log_dir                = args.tensorboard_dir,
             histogram_freq         = 0,
             batch_size             = args.batch_size,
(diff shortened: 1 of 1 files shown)


keras_retinanet/bin/train.py
```diff
@@ -109,14 +109,31 @@ def create_callbacks(model, training_model, prediction_model, validation_generat
         checkpoint = RedirectModel(checkpoint, prediction_model)
         callbacks.append(checkpoint)
 
+    if args.tensorboard_dir:
+        tensorb = keras.callbacks.TensorBoard(
```
hgaiser: I prefer the name `tensorboard` instead.

keras_retinanet/bin/train.py
```diff
@@ -109,14 +109,31 @@ def create_callbacks(model, training_model, prediction_model, validation_generat
...
+
     if args.evaluation and validation_generator:
         if args.dataset_type == 'coco':
             from ..callbacks.coco import CocoEval
 
             # use prediction model for evaluation
             evaluation = CocoEval(validation_generator)
         else:
-            evaluation = Evaluate(validation_generator)
+            if args.tensorboard_dir:
```
hgaiser: Hmm what about always initializing a `tensorboard` variable, but leave it `None` if `args.tensorboard_dir` is not set. That way this line can remain (without if):

```python
evaluation = Evaluate(validation_generator, tensorboard=tensorboard)
```

keras_retinanet/bin/train.py
awilliamson: Or perhaps `tb_callback` to more closely reflect what the Object is?

de-vri-es: I'd prefer `tensorboard_callback` then. `tb` is a bit cryptic to me.

mxvs:
Review remarks have been fixed, I went with the naming suggestion offered by @de-vri-es

hgaiser:
LGTM, thanks!

PR #295, opened 2018-02-23
Specified python 3.6

wassname:
I get an import error in python3.5 and you guys have travis set up for 3.6, so I though I would add it to the readme.


hgaiser:
Thanks! What was the import error by the way ?

wassname:
Reffering to https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/models/resnet.py#L47 it was something like "module 'keras.applications' has no attribute 'imagenet_utils'" from keras-retinanet/keras_retinanet/models/resnet.py.

So it could be fixed for python3.5 with a `import keras.applications.imagenet_utils` (or maybe with futures or something) but I figure you guys have enough work supporting python3.6.

PR #296, opened 2018-02-23
Change default nms-threshold as 0.5

mkocabas:
I have only changed the nms layer default nms-threshold parameter.


hgaiser:
Thanks!

For history purposes: the NMS threshold in the paper was 0.5, I don't know why we had 0.4 to be honest.

PR #297, opened 2018-02-23
Oid add filter option

lvaleriu:


hgaiser:
To be honest, I don't know much about OID to say if this is sensible or not ;) but I trust that you do.

One small thing, could you make the optional arguments use dashes instead of underscores? So `--fixed-labels` instead of `--fixed_labels` (and other arguments too).

lvaleriu:
Done!

As for OID: since this dataset is big, there is a common habit to train only on a subset. This subset is chosen by selecting a list of labels/sublabels. So label filtering is an interesting option.

keras_retinanet/bin/debug.py
```diff
@@ -108,9 +108,9 @@ def csv_list(string):
     oid_parser.add_argument('main_dir', help='Path to dataset directory.')
     oid_parser.add_argument('subset', help='Argument for loading a subset from train/validation/test.')
     oid_parser.add_argument('--version',  help='The current dataset version is V3.', default='2017_11')
-    oid_parser.add_argument('--labels_filter',  help='A list of labels to filter.', type=csv_list, default=None)
-    oid_parser.add_argument('--annotation_cache_dir', help='Path to store annotation cache.', default='.')
-    oid_parser.add_argument('--fixed_labels', help='Use the exact specified labels.', default=False)
+    oid_parser.add_argument('--labels-filter',  help='A list of labels to filter.', type=csv_list, default=None)
+    oid_parser.add_argument('--annotation-cache_dir', help='Path to store annotation cache.', default='.')
```
hgaiser: Forgot one here.

hgaiser:
I see, thanks for the PR. Seems you didn't change one `--annotation_cache_dir` correctly, if you change that, I'll merge it.

lvaleriu:
Damn ... I didnt see this one. It's fixed now. Thanks.

keras_retinanet/bin/debug.py
lvaleriu: Fixed

PR #302, opened 2018-02-24
typo blue should be red in debug

wassname:
A minor typo where the comments refers to red but it's plotted as blue


hgaiser:
How are you visualizing the image? OpenCV uses the BGR notation and we do too (since most systems seem to use that notation). However, matplotlib (probably rightfully) uses the RGB notation. If you visualised using matplotlib you might get the colorspaces wrong.

wassname:
Ah that's it, I was visualising using matplotlib, sorry about that!

PR #305, opened 2018-02-26
update feature pyramid branch to consist with original paper and Detectron

xmyqsh:
update feature pyramid branch to consist with original paper and [Detectron](https://github.com/facebookresearch/Detectron/blob/master/lib/modeling/FPN.py#L116-L224)


hgaiser:
Thanks for the PR. I had noticed this difference too, but wanted to investigate it a bit further (and retrain the pretrained network) before merging. The version I'm using to test this is: https://github.com/fizyr/keras-retinanet/tree/fpn-correction?files=1

If you don't mind I'd like to close this issue, run the remainder of experiments on that branch and then make a PR for that. Nevertheless, good eye!

PR #307, opened 2018-02-26
Add multi-scale evaluation option

mkocabas:
This PR adds multi-scale evaluation option only to `evaluate_coco.py` script. Multi scale evaluation can be useful for users who need more accurate results. Default value of this option is set to false to avoid slow evaluation with default parameters. Scales are fixed and `[(688, 1145), (800, 1333), (1200, 1600)]` which gives ~0.06 mAP boost on COCO. Users can change scales manually according to their needs.


hgaiser:
First of all, thanks for the contribution. I've thought about this for a while, whether it's something we want in this implementation, and I've come to the conclusion that it's better to keep things simpler and leave this out. This PR adds quite a bit of duplicated code and complexity in the evaluation code. Also, it is only implemented for COCO and there's no option for adjusting the scales through arguments. Lastly, I believe this is not something that generally appears to be done in object detection research (or is it?).

I understand where you're coming from, but I prefer to keep the code as "lean" as possible. Of course people using keras-retinanet are free to use as many scales as they like, but that will be their responsibility.

PR #308, opened 2018-02-26
Implement new backbone => DenseNet

vidosits:
Hi!

I've added (and tested on my end) a new backbone: DenseNets from keras.applications. The PR is based on this PR: Implement new backbone => MobileNet #286. I think it's quite trivial (with one exception) since the DenseNet architectures in keras.applications all have 4 stages. So we take the last layer of every stage/block as the given level of the feature pyramid. The only thing I wasn't certain about whether to take the last layer of the block (which is a concatenation) or the last conv layer of a given block, but trying both it looks like it's better to take the very last layer of each block (the concat one). Feel free to edit it if you feel like it.


keras_retinanet/models/densenet.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+
+import keras
+from keras.applications.densenet import DenseNet, get_file
+
+from ..models import retinanet
+
+BASE_WEIGHT_PATH = 'https://github.com/fchollet/deep-learning-models/releases/download/v0.8/'
+
+custom_objects = retinanet.custom_objects.copy()
```
hgaiser: No copy is required here.

keras_retinanet/models/densenet.py
```diff
@@ -0,0 +1,77 @@
...
+
+allowed_backbones = {'densenet121': [6, 12, 24, 16], 'densenet169': [6, 12, 32, 32], 'densenet201': [6, 12, 48, 32]}
+
+
+def download_imagenet(backbone):
+    """ Download pre-trained weights for the specified backbone name. This name is in the format
+        {backbone}_weights_tf_dim_ordering_tf_kernels_notop where backbone is the densenet + number of layers (e.g. densenet121).
+        For more info check the explanation from the keras densenet script itself
```
hgaiser: Maybe it'd be nice to add a link to that script here.

keras_retinanet/models/densenet.py
```diff
@@ -0,0 +1,77 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import keras
+from keras.applications.densenet import DenseNet, get_file
+
+from ..models import retinanet
+
+BASE_WEIGHT_PATH = 'https://github.com/fchollet/deep-learning-models/releases/download/v0.8/'
```
hgaiser: I prefer this to be :

```python
WEIGHT_PATH = 'https://github.com/fchollet/deep-learning-models/releases/download/v0.8/densenet{}_weights_tf_dim_ordering_tf_kernels_notop.h5'
```

So that later on we can do:

```python
weights_url = WEIGHT_PATH.format(backbone)
```

keras_retinanet/models/densenet.py
```diff
@@ -0,0 +1,77 @@
...
+def validate_backbone(backbone):
+    """ Validate the backbone choice.
+    # Arguments
+        backbone    : Backbone name.
+    """
+
+    backbone = backbone.split('_')[0]
+
+    if backbone not in allowed_backbones:
+        raise ValueError('Backbone (\'{}\') not in allowed backbones ({}).'.format(backbone, allowed_backbones))
```
hgaiser: Print `allowed_backbones.keys()` instead.

keras_retinanet/models/densenet.py
```diff
@@ -0,0 +1,77 @@
...
+    return weights_path
+
+
+def validate_backbone(backbone):
+    """ Validate the backbone choice.
+    # Arguments
+        backbone    : Backbone name.
+    """
+
+    backbone = backbone.split('_')[0]
```
hgaiser: What is the `_` for?

panda9095:
Just tried the densenet_backbone PR, but got an _Incompatible shape_ error at Line 82 in losses.py.

vidosits:
Thats very interesting, let me check it too.

keras_retinanet/models/densenet.py
vidosits: you are correct! my bad

keras_retinanet/models/densenet.py
vidosits: added.

vidosits:
@panda9095 I'm pretty sure this is caused by not using input images with width:height ratio of 1, this is a problem with the backbone, although I'm not sure what causes it. Maybe @hgaiser can shed some light to this. What IS sure is that the unit tests all pass when the input images have w:h ratio of 1, but fail otherwise.

@hgaiser I also committed the changes you requested.

vidosits:
image
If you look at test_densenet.py you can see that I'm using a square ratio image, thats why the test passes.

lvaleriu:
This pull request: https://github.com/fizyr/keras-retinanet/pull/298 fixes this.

vidosits:
> This pull request: #298 fixes this.

@lvaleriu Is there anything I should do in this PR? Also I'm not quite sure why the Travis build times out...

hgaiser:
> This pull request: #298 fixes this.

I'm not sure yet if that PR has the best approach for this problem, I'll have to look into this soon so these PRs can be merged.

lvaleriu:
@vidosits You might want to remove all backbones from the unit test and choose only the smallest. The travis instance limits execution of an unit test to 10 minutes and they dont seem to have gpu-us.

@hgaiser Yes, a solution is needed since i got the same thing when trying to integrate ResNeXt or InceptionResnetV2 as backends. This PR is a solution though.

vidosits:
@lvaleriu thanks, kept only densenet121, decreased the input image size to imagenet 224x224 too, hopefully travis won't time out this time

hgaiser:
Does this backbone work without having to find out the pyramid shapes from the model, as in #298 ?

vidosits:
@hgaiser I'm not exactly sure what #298 does, but right now the densenet121/161/169 backbones can only be used with images of height:width ratio of 1.

hgaiser:
@vidosits yeah that sounds like something addressed in #298 . I can't merge that yet, so we'll have to wait until that is merged before this can be merged.

vidosits:
@hgaiser okay, thank you! very much appreciate the feedback :)

hgaiser:
#298 is merged, could you update this PR and test if DenseNet is working as intended?

vidosits:
@hgaiser sure, on it

vidosits:
@hgaiser updated it, backbone fails for every image with height:width ratio different than h:w = 1:1, even after #298, according to the generator the regression output for a batch with size [1, 200, 400, 3] should be [1,15318,4], but the actual output of the model is of shape [1,14814,4].
The same model runs with input batch of size [1, 224, 224, 3], the regression output is of size [1, 9441, 5]. I'm sure that I'm the one messing up, but I'm not sure where it goes wrong.

hgaiser:
Hmm #298 contains tools to find out the correct number of anchors for a model, but by default it will use the "old" behaviour. I think you need to do something similar to this: https://github.com/fizyr/keras-retinanet/pull/298/files#diff-fb5be6ca95a3190ebad5eaedf0a5ce65R367

vidosits:
@hgaiser , @lvaleriu thanks, I completely missed that part of #298 , fixed it in PR (added it to train.py in case of using the densenet backbones)

hgaiser:
Nice, thanks for this!

vidosits:
hey, no problem, my pleasure!

Issue #171, opened 2017-12-07
bbox_transform formula


Cre1th:
Hey, I was wondering why 1.0 is added to the widths and heights in [bbox_transform](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/utils/anchors.py).

```
anchor_widths  = anchors[:, 2] - anchors[:, 0] + 1.0
anchor_heights = anchors[:, 3] - anchors[:, 1] + 1.0
anchor_ctr_x   = anchors[:, 0] + 0.5 * anchor_widths
anchor_ctr_y   = anchors[:, 1] + 0.5 * anchor_heights
```
...
```
targets_dx = (gt_ctr_x - anchor_ctr_x) / anchor_widths
targets_dy = (gt_ctr_y - anchor_ctr_y) / anchor_heights
targets_dw = np.log(gt_widths / anchor_widths)
targets_dh = np.log(gt_heights / anchor_heights)
```

It seems to me the purpose is to avoid division by zero. But I think, taking the maximum(1, anchor_widths/anchor_heights) is better suited for this task.
Especially because I think this introduces a slight error. Let's say the center offset in `x` is `10` then `target_dx = 10 / (anchors[:, 2] - anchors[:, 0] + 1.0)`. This is the value we want our regressor to have.
Using now [inv_bbox_transform](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/backend/common.py) even with this target regression results in
`pred_ctr_x = ctr_x + target_dx * widths` which does not yield the result we want of `ctr_x + 10`. This is because `target_dw * widths` is slightly less then 10 as in [inv_bbox_transform](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/backend/common.py) the + 1.0 is not taken into account see `widths  = boxes[:, :, 2] - boxes[:, :, 0]`.

So my question is: Does this +1.0 serve any other purpose?

hgaiser:
Good question! The short and boring answer is, that's how it is implemented in [py-faster-rcnn](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/fast_rcnn/bbox_transform.py). However in their inverse transformation they also use `+ 1`'s.

I think I removed them when implementing the tensorflow implementation of the inverse transformation but never checked this part. In my opinion, x2 and y2 are not inclusive so the width is simply `x2 - x1`.

That being said, I think you're right! Thanks for bringing this to my attention, I'll see about fixing that.

[...]

PR #311, opened 2018-03-02
Improve COCO results (FPN + NMS changes).

hgaiser:
This PR does the following:

- Correct FPN computation as per their paper.
- Correct bbox transform formula (https://github.com/fizyr/keras-retinanet/issues/171).
- Change default image size to 800px.
- Change NMS to do per class NMS.

Since this is a breaking change for all models trained before this PR, I propose to tag `master` as 0.2 after merging and publishing a new pretrained model.


keras_retinanet/utils/anchors.py
```diff
@@ -171,16 +171,20 @@ def bbox_transform(anchors, gt_boxes, mean=None, std=None):
     elif not isinstance(std, np.ndarray):
         raise ValueError('Expected std to be a np.ndarray, list or tuple. Received: {}'.format(type(std)))
 
-    anchor_widths  = anchors[:, 2] - anchors[:, 0] + 1.0
-    anchor_heights = anchors[:, 3] - anchors[:, 1] + 1.0
+    anchor_widths  = anchors[:, 2] - anchors[:, 0]
```
awilliamson: I thought we needed these +1.0s due to dataset reasons? Can't fully recall, but I remember being told something along those lines?

keras_retinanet/models/retinanet.py
```diff
@@ -103,14 +103,15 @@ def default_regression_model(num_anchors, pyramid_feature_size=256, regression_f
 
 def __create_pyramid_features(C3, C4, C5, feature_size=256):
     # upsample C5 to get P5 from the FPN paper
-    P5           = keras.layers.Conv2D(feature_size, kernel_size=1, strides=1, padding='same', name='P5')(C5)
+    P5           = keras.layers.Conv2D(feature_size, kernel_size=1, strides=1, padding='same', name='C5_reduced')(C5)
```
awilliamson: Verified this is in line with Faster RCNN paper and their Pyramid outline, including the exceptions noted in the Focal Loss paper. This was done at the time @hgaiser was writing.

keras_retinanet/utils/anchors.py
hgaiser: Those +1's are from `py-faster-rcnn` originally. They encode a box as `[x1, x2]`, `[y1, y2]`, we encode it as `[x1, x2)` and `[y1, y2)`. That means we should compute the width as `x2 - x1` and height as `y2 - y1`. `py-faster-rcnn` (and Detectron for that matter) computes it as `x2 - x1 + 1` and `y2 - y1 + 1`.

Somewhere a while back we switched to this new notation, which in my opinion makes much more sense. However this code was left unchanged.

PR #330, opened 2018-03-13
Add tensorboard mAP for COCO

jjiunlin:
Small change but maybe useful.
It can log mAP for coco dataset if `--tensorboard-dir` enable.

### DIFF
--- a/keras_retinanet/utils/coco_eval.py
+++ b/keras_retinanet/utils/coco_eval.py
@@ -69,7 +69,7 @@ def evaluate_coco(generator, model, threshold=0.05):
         print('{}/{}'.format(index, generator.size()), end='\r')
 
     if not len(results):
-        return [0]
+        return
 
     # write output
     json.dump(results, open('{}_bbox_results.json'.format(generator.set_name), 'w'), indent=4)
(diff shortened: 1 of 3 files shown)


keras_retinanet/utils/coco_eval.py
```diff
@@ -69,7 +69,7 @@ def evaluate_coco(generator, model, threshold=0.05):
         print('{}/{}'.format(index, generator.size()), end='\r')
 
     if not len(results):
-        return
+        return [0]
```
de-vri-es: Why `return [0]`? Wouldn't `return []` make more sense?

jjiunlin: When I test with step=1 epoch=1 training from scratch.There are no results.I can't get mAP list from`coco_eval.stats`.The original `return` raise `TypeError: 'NoneType' object is not subscriptable`. So I assign `[0]` for mAP=0 for there are no results. 
 
`return []` will raise `IndexError: list index out of range` because i write  `summary_value.simple_value = mAP[0]`in `coco.py line35`.

hgaiser: In the success path, does `mAP` contain all values of the summary? Maybe it's better to iterate over the `mAP` list and add them all? Is there a way to map them to the correct name of the value?

jjiunlin:
@hgaiser I add all values of `coco_eval.stats` with correct tag to tensorboard.It looks like this.
Also I revise to the original `return` .If `return None`, it won't write values to tensorboard.

hgaiser:
Awesome, thanks!

PR #333, opened 2018-03-14
Output FPN and docs.

hgaiser:
This PR adds a flag to the retinanet model to output FPN features. This is necessary for `keras-maskrcnn` because it needs these features to compute a mask.

I also accidentally added doc strings to `retinanet.py`.

### DIFF
--- a/keras_retinanet/layers/_misc.py
+++ b/keras_retinanet/layers/_misc.py
@@ -174,7 +174,10 @@ class RegressBoxes(keras.layers.Layer):
         return input_shape[0]
 
     def get_config(self):
-        return {
+        config = super(RegressBoxes, self).get_config()
+        config.update({
             'mean': self.mean.tolist(),
             'std' : self.std.tolist(),
-        }
+        })
+
+        return config
--- a/keras_retinanet/models/retinanet.py
+++ b/keras_retinanet/models/retinanet.py
@@ -21,6 +21,9 @@ from .. import losses
 
 import numpy as np
 
+"""
+A dictionary mapping custom layer names to the correct classes.
+"""
 custom_objects = {
     'UpsampleLike'          : layers.UpsampleLike,
     'PriorProbability'      : initializers.PriorProbability,
@@ -40,6 +43,18 @@ def default_classification_model(
     classification_feature_size=256,
     name='classification_submodel'
 ):
+    """ Creates the default regression submodel.
+
+    Args
+        num_classes                 : Number of classes to predict a score for at each feature level.
+        num_anchors                 : Number of anchors to predict classification scores for at each feature level.
+        pyramid_feature_size        : The number of filters to expect from the feature pyramid levels.
+        classification_feature_size : The number of filters to use in the layers in the classification submodel.
+        name                        : The name of the submodel.
+
+    Returns
+        A keras.models.Model that predicts classes for each anchor.
+    """
     options = {
         'kernel_size' : 3,
         'strides'     : 1,
@@ -74,6 +89,17 @@ def default_classification_model(
 
 
 def default_regression_model(num_anchors, pyramid_feature_size=256, regression_feature_size=256, name='regression_submodel'):
+    """ Creates the default regression submodel.
+
+    Args
+        num_anchors             : Number of anchors to regress for each feature level.
+        pyramid_feature_size    : The number of filters to expect from the feature pyramid levels.
+        regression_feature_size : The number of filters to use in the layers in the regression submodel.
+        name                    : The name of the submodel.
+
+    Returns
+        A keras.models.Model that predicts regression values for each anchor.
+    """
     # All new conv layers except the final one in the
     # RetinaNet (classification) subnets are initialized
     # with bias b = 0 and a Gaussian weight fill with stddev = 0.01.
@@ -102,6 +128,17 @@ def default_regression_model(num_anchors, pyramid_feature_size=256, regression_f
 
 
 def __create_pyramid_features(C3, C4, C5, feature_size=256):
+    """ Creates the FPN layers on top of the backbone features.
+
+    Args
+        C3           : Feature stage C3 from the backbone.
+        C4           : Feature stage C4 from the backbone.
+        C5           : Feature stage C5 from the backbone.
+        feature_size : The feature size to use for the resulting feature levels.
+
+    Returns
+        A list of feature levels [P3, P4, P5, P6, P7].
+    """
     # upsample C5 to get P5 from the FPN paper
     P5           = keras.layers.Conv2D(feature_size, kernel_size=1, strides=1, padding='same', name='C5_reduced')(C5)
     P5_upsampled = layers.UpsampleLike(name='P5_upsampled')([P5, C4])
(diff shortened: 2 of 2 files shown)


keras_retinanet/models/retinanet.py
```diff
@@ -102,6 +128,17 @@ def default_regression_model(num_anchors, pyramid_feature_size=256, regression_f
...
+    """ Creates the FPN layers on top of the backbone features.
+
+    Args
+        C3           : Feature stage C3 from the backbone.
+        C4           : Feature stage C4 from the backbone.
+        C5           : Feature stage C5 from the backbone.
+        feature_size : The feature size to use for the resulting feature levels.
+
+    Returns
+        A list of feature levels (P3, P4, P5, P6, P7).
```
awilliamson: Perhaps change the comment to use list notation rather than tuple notation, just in case this could be misconstrued.

keras_retinanet/models/retinanet.py
```diff
@@ -148,31 +195,78 @@ def num_anchors(self):
...
+        A list of tuple, where the first element is the name of the submodel and the second element is the submodel itself.
+    """
     return [
         ('regression', default_regression_model(anchor_parameters.num_anchors())),
         ('classification', default_classification_model(num_classes, anchor_parameters.num_anchors()))
     ]
 
 
 def __build_model_pyramid(name, model, features):
+    """ Builds pyramid for a single submodel.
```
awilliamson: More accurately applies the submodel to each FPN level. I wouldn't necessarily call this building a pyramid or constructing it, merely applying over an existing pyramid of features. Perhaps the function signature is misleading?

keras_retinanet/models/retinanet.py
```diff
@@ -193,28 +311,71 @@ def retinanet(
...
+
+    Args
+        inputs      : keras.layers.Input (or list of) for the input to the model.
+        num_classes : Number of classes to classify.
+        output_fpn  : If True, the last 5 outputs before detections are P3...P7.
+        name        : Name of the model.
+        *kwargs     : Additional kwargs to pass to the minimal retinanet model.
+
+    Returns
+        Model with inputs as input and as output the output of each submodel for each pyramid level and the detections.
```
awilliamson: Sounds a little strange when reading.

keras_retinanet/models/retinanet.py
```diff
@@ -193,28 +311,71 @@ def retinanet(
...
+    """
+    model = retinanet(inputs=inputs, num_classes=num_classes, output_fpn=output_fpn, **kwargs)
 
     # we expect the anchors, regression and classification values as first output
     anchors        = model.outputs[0]
     regression     = model.outputs[1]
     classification = model.outputs[2]
 
+    if output_fpn:
+        other = model.outputs[3:-5]
```
awilliamson: Perhaps more clarification on what other constitutes? Clearly relates to having output_fpn set to True. Maybe a comment just above L362?

keras_retinanet/layers/_misc.py
```diff
@@ -174,7 +174,10 @@ def compute_output_shape(self, input_shape):
         return input_shape[0]
 
     def get_config(self):
-        return {
+        config = super(RegressBoxes, self).get_config()
```
awilliamson: This doesn't seem to fit with the feature theme of FPN split output w/ Documentation.

keras_retinanet/models/retinanet.py
hgaiser: :+1:

keras_retinanet/models/retinanet.py
hgaiser: :+1:

keras_retinanet/models/retinanet.py
hgaiser: Re-worded (?) it.

keras_retinanet/models/retinanet.py
hgaiser: :+1:

keras_retinanet/layers/_misc.py
hgaiser: True, and it doesn't break anything, but it was technically incorrect. Didn't feel it deserved its own PR but maybe I should have just pushed it to master directly instead.

hgaiser:
Updated the docs, thanks for the review! Awaiting CI, will merge after.

keras_retinanet/models/retinanet.py
```diff
@@ -148,31 +195,76 @@ def num_anchors(self):
...
+        layers.Anchors(
             size=anchor_parameters.sizes[i],
             stride=anchor_parameters.strides[i],
             ratios=anchor_parameters.ratios,
             scales=anchor_parameters.scales,
             name='anchors_{}'.format(i)
-        )(f))
-    return keras.layers.Concatenate(axis=1)(anchors)
+        )(f) for i, f in enumerate(features)
+    ]
```
de-vri-es: While I like doing things in single expressions, I'm not so sure the list comprehension is really an improvement over the loop in terms of readability.

keras_retinanet/models/retinanet.py
```diff
@@ -193,28 +309,72 @@ def retinanet(
...
 
     # additionally apply non maximum suppression
     if nms:
         detections = layers.NonMaximumSuppression(name='nms')([boxes, classification] + model.outputs[3:])
     else:
-        detections = keras.layers.Concatenate(axis=2)([boxes, classification] + model.outputs[3:])
+        detections = keras.layers.Concatenate(axis=2, name='detections')([boxes, classification] + model.outputs[3:])
+
+    # construct list of outputs
+    outputs = [regression, classification] + other + (fpn if output_fpn else []) + [detections]
```
de-vri-es: This is the only place I see `other` and `fpn` being used, and they are being concatenated. Do we really need to distinguish `output_fpn` and `not output_fpn` then? It looks like `other = model.outputs[3:]` would suffice in both cases (in which case I would also replace all occurrences of `model.outputs[3:]` with `other`). If so, output_fpn can also be caught by **kwargs.

If `output_fpn` can't be removed, how about just setting `fpn = []` in the `not output_fpn` case above, and simplify this to `other + fpn`?

keras_retinanet/models/retinanet.py
```diff
@@ -193,28 +309,72 @@ def retinanet(
     features = create_pyramid_features(C3, C4, C5)
 
     # for all pyramid levels, run available submodels
-    pyramid = __build_pyramid(submodels, features)
-    anchors = __build_anchors(anchor_parameters, features)
+    pyramids = __build_pyramid(submodels, features)
+    anchors   = __build_anchors(anchor_parameters, features)
```
de-vri-es: Spacing error?

keras_retinanet/models/retinanet.py
hgaiser: Ah actually that was a boo-boo which I already fixed, but didn't push yet. `other` should've been used instead of `model.outputs[3:]` (which it does now). The thing is that the last 5 elements are the FPN features if `output_fpn` is used, so `model.outputs[3:]` wouldn't work in every case.

keras_retinanet/models/retinanet.py
hgaiser: Yup.

hgaiser:
> What about looking up the layers by name in mask RCNN? Is that a viable option?

Yeah that's an option. Not sure which one would be "better" though. You think looking up the layers is better I suppose?

hgaiser:
Okay okay, I removed `output_fpn` again, you're right. It is trivial to lookup these values in `keras-maskrcnn`. This PR only includes doc and some minor changes then. Will merge after CI approves.

keras_retinanet/models/retinanet.py
awilliamson: I personally like list comprehensions, but it may reduce readability for some users. Although I do prefer not having an empty definition for anchors, it's all neatly packaged via list comp.

PR #335, opened 2018-03-16
draw_detections (in utils.visualization) fix, failed when not supplying a generator

srslynow:
```
caption = (generator.label_to_name(label) if generator else label) + ': {0:.2f}'.format(score)
```

Fails when not using a generator: the ```+``` operator doesn't work on ```label```, which is a float object


hgaiser:
Thanks!

PR #337, opened 2018-03-16
Label colors

hgaiser:
This PR adds a utility function to get colors for labels. Labels until 80 have currently been assigned random colors, other labels will get the color green.


keras_retinanet/utils/colors.py
```diff
@@ -0,0 +1,113 @@
+import numpy as np
+import matplotlib.colors
```
de-vri-es: I assume you added the list below to avoid depending on `matplotlib`, but you forgot to remove this import? `numpy` doesn't seem to be necessary anymore either.

de-vri-es:
One in-line comment, looks good other than that.

Shouldn't this be used in `evaluate.py` though?

keras_retinanet/utils/colors.py
hgaiser: Very perceptive mister @de-vri-es , I removed those.

hgaiser:
> Shouldn't this be used in evaluate.py though?

Yep, I added it there now too. Thanks!

PR #339, opened 2018-03-16
Fixup loss normalization

hgaiser:
After careful inspection of Detectron, it appears our normalization of losses was slightly incorrect. We were normalizing each batch with the number of positive anchors in that batch, but every value should be divided by the total number of positive anchors. This PR fixes that and adds a unit test for the smooth L1 loss.

### DIFF
--- a/keras_retinanet/losses.py
+++ b/keras_retinanet/losses.py
@@ -23,19 +23,6 @@ def focal(alpha=0.25, gamma=2.0):
         labels         = y_true
         classification = y_pred
 
-        # compute the divisor: for each image in the batch, we want the number of positive anchors
-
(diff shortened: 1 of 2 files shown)


huatangzhi:
have this affect a lot in mAP？

hgaiser:
> have this affect a lot in mAP？

Probably not. It might affect training with batch_size > 1 though.

de-vri-es:
Assuming it is tested, I have no comments. Code looks good, though I didn't read up on the theory.

A test training + evaluation would definitely be prudent to confirm there is no performance loss though.

hgaiser:
I tested it briefly due to resource limitations, but it was perfectly normal.

PR #348, opened 2018-03-22
refactor import of Image submodule

DiegoAgher:
# What does this PR do?
Fix the conflict that arises from using the `Image` submodule of `pillow` package.
This issue is caused by the newest version of pillow, `5.0.0`.

### DIFF
--- a/keras_retinanet/utils/image.py
+++ b/keras_retinanet/utils/image.py
@@ -20,13 +20,13 @@ import time
 import numpy as np
 import scipy.ndimage as ndi
 import cv2
-import PIL
+from PIL import Image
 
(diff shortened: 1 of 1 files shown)


de-vri-es:
Can you describe the issue that arises without this PR in more detail, maybe with a stack trace?

Looking at the code alone, both versions seem like they should work, or neither should work.

de-vri-es:
Ah, nevermind, I could easily reproduce the problem locally.

Will merge after CI passes. Thanks for the PR :)

DiegoAgher:
cool, anytime!
Really nice repo, congrats!

Issue #345, opened 2018-03-21
create_callbacks() crash under Python 2.7 if snapshots


pderian:
In `train.py:create_calbacks()` l. 99: `exist_ok` is valid option of `os.makedirs()` under Python 3, not Python 2.7. Tests pass as they have the` --no-snapshot` option.

https://github.com/fizyr/keras-retinanet/blob/7ba5ec2bf3497c9fd4c8dd2534d0d1d15e717a9b/keras_retinanet/bin/train.py#L99

hgaiser:
Nice catch! Can you make a PR to fix this? Probably it's easiest to wrap it in an `if os.path.isdir(...)`

de-vri-es:
The proper way to deal with existing directories is to just try to create it and catch the error. Checking and then creating has a small race condition.

hgaiser:
Fair enough, let's go that way.

de-vri-es:
To be 100% correct we should check the file type if it exists already. If an error is raised and the path is not a directory, the error shouldn't be ignored.

pderian:
Something along the lines of:
```
    # save the prediction model
    if args.snapshots:
        # ensure directory created first; otherwise h5py will error after epoch.
        try:
            os.makedirs(args.snapshot_path)
        # if an error is raised
        except OSError as err_makedirs:
            # and it is NOT because the dir exists already
            if not os.path.isdir(args.snapshot_path):
                raise(err_makedirs)
            # otherwise we're good (following the previous behavior)
```
That is to say: if `os.makedirs()` failed because the dir exists already, it is fine (mimics `exist_ok=True`), but if it failed and `args.snapshot_path` is not a dir then the error propagates.

de-vri-es:
Pretty much. One thing though: the best way to re-raise a caught exception is to simply invoke `raise` without arguments. That way the stack trace of the original exception is preserved (though the new raise location is also added, I think).

It would also be nice to add it as a utility function. It is fine to define it locally in the file where we need it though. If we need it in other places too we can move it to `utils` later.

de-vri-es:
Fixed by #350. Thanks again for your contribution :)

PR #350, opened 2018-03-23
fixed makedirs crash - issue #345

pderian:
`os.makedirs()` with `exist_ok` option is not valid under Python 2.7 [(Issue #345)](https://github.com/fizyr/keras-retinanet/issues/345).
Proposed fix is compatible with 2.7 while preserving former behavior, i.e:
- try to create directory tree;
- upon failure, an OSError is raised by os.makedirs;
- if this is because the directory exsits already, the error passes; otherwise it is raised.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -44,6 +44,16 @@ from ..utils.transform import random_transform_generator
 from ..utils.keras_version import check_keras_version
 from ..utils.model import freeze as freeze_model
 
+def makedirs(path):
+    # Intended behavior: try to create the directory,
+    # pass if the directory exists already, fails otherwise.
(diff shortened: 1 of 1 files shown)


keras_retinanet/bin/train.py
```diff
@@ -45,6 +45,17 @@
 from ..utils.model import freeze as freeze_model
 
 
+def makedirs(path):
+    # Intended behavior: try to create the directory,
+    # pass if the directory exists already, fails otherwise.
+    # Meant for Python 2.7/3.n compatibility.
+    try:
+        os.makedirs(path)
+    except OSError as err_makedirs:
```
de-vri-es: No need to give the error a name since we're just invoking `raise` without arguments. Other than that, looks good :+1:

PR #353, opened 2018-03-26
Fix debug script

eduramiba:
Fix a small error on debug script


de-vri-es:
Right, this was changed by #298.

Thanks for the fix!

PR #357, opened 2018-03-27
Update visualization.py

pedroconceicao:
"image" argument missing when calling draw_box()

### DIFF
--- a/keras_retinanet/utils/visualization.py
+++ b/keras_retinanet/utils/visualization.py
@@ -75,7 +75,7 @@ def draw_detections(image, boxes, classification, color=None, generator=None, sc
 
     for i, label in np.transpose(selection):
         c = color if color is not None else label_color(label)
-        draw_box(boxes[i, :], color=c)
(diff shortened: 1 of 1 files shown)


hgaiser:
LGTM, waiting for Travis and will merge afterwards.

hgaiser:
That last commit removes the label specific coloring ability.

pedroconceicao:
How so? 

I tried the script that I submitted and I get the same results as before because in the for loop the bounding boxes are overwritten. When calling drawboxes() black boxes are drawn and then again in the foor loop colored boxes are drawn exaclty on top of the black ones. 

Why don't just draw the colored ones? You get the same output with or without drawboxes().

hgaiser:
Ah nevermind, I misinterpreted the diff. Could you fix the pep8 error ?

pedroconceicao:
Yes, I did. I did a clean install, trained again and the evaluation worked. Thank you!

PR #368, opened 2018-04-04
Much faster method to draw the bounding boxes

rodrigo2019:
For loop's in Python are too slow, I used this method to use  a loop as short as possible to speed up the drawing process, with the original method the time to draw was running around 0.4s per image, with new method the time is now around 0.01s
It is usefull if want to use the network in a video.


hgaiser:
Thanks! I see where you're going with this, but the method you use looks a bit complicated. If you change the PR to match [this](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/utils/coco_eval.py#L47) method I will gladly merge it.

examples/ResNet50RetinaNet.ipynb
```diff
@@ -153,17 +153,17 @@
     "boxes /= scale\n",
     "\n",
     "# visualize detections\n",
-    "for idx, (label, score) in enumerate(zip(predicted_labels, scores)):\n",
-    "    if score < 0.5:\n",
-    "        continue\n",
-    "        \n",
-    "    color = label_color(label)\n",
+    "for i in np.where(scores > 0.5)[0]:\n",
```
hgaiser: It would be easier to loop over `nms_classification` directly:
```python
for i, label in np.transpose(np.where(nms_classification > 0.5)):
    b = boxes[i, :].astype(int)
    ...
```

Then you can delete some of the variables as well (scores and predicted_labels).

rodrigo2019: Ok, I will change, but I kept in that way for those which are learning the code (like me) make a little bit easier to understand.
Thank you.

rodrigo2019:
@hgaiser could you check your method? your change makes the method take around 54ms/image, and the method from my last commit take around 18ms/image.

Thank you for sharing your knowledge.

hgaiser:
You're absolutely correct. I just did some measurements and got more or less the same results as you. Although I do think the proposed method is clearer (requires fewer variables), the difference in time is too significant.

The two methods aren't the same however. The proposed method would allow for one box to have multiple classes assigned to it. The method from this PR picks the class with the highest score. That's not necessarily a bad thing, but it's something to take into account. For evaluation on COCO for instance, we want one box to possibly be multiple classes, so we have to use the version I proposed there.

I'm okay with merging this. Is there anything you wanted to change or is it done?

rodrigo2019:
I didn't know about the coco evaluation, but this PR keeps the same logic as before, correct? So I'm okay, and it's done.
Thanks a lot for your help.

hgaiser:
Thank you for the PR!

It also showed me that the current implementation of NMS is very inefficient (though elegant ;)). I'm working on a refactor that will remove the sparsity of the NMS output.

Programmerwyl:
@rodrigo2019  Can you tell me how you handle video with retinanet ？
Thank you very much

rodrigo2019:
Using this tutorial applying the retinanet frame by frame using cv2.VideoCapture,
you can check [here](https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_gui/py_video_display/py_video_display.html) how to use it.

PR #376, opened 2018-04-07
delete clipnorm

luliyucoordinate:
delete clipnorm

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ model.compile(
         'regression'    : keras_retinanet.losses.smooth_l1(),
         'classification': keras_retinanet.losses.focal()
     },
-    optimizer=keras.optimizers.adam(lr=1e-5, clipnorm=0.001)
+    optimizer=keras.optimizers.adam(lr=1e-5)
 )
 ```
 2) Create generators for training and testing data (an example is show in [`keras_retinanet.preprocessing.PascalVocGenerator`](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/preprocessing/pascal_voc.py)).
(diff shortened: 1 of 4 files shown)


hgaiser:
Could you merge all PRs in one PR? They seem highly correlated.

Also, why should we delete this?

luliyucoordinate:
Sorry l don't know who to merge all in Browser interface.
The reason for deleting it is not stated in keras Documentation
```
Adam
keras.optimizers.Adam(lr=0.001, beta_1=0.9, beta_2=0.999, epsilon=None, decay=0.0, amsgrad=False)
Adam optimizer.

Default parameters follow those provided in the original paper.

Arguments

lr: float >= 0. Learning rate.
beta_1: float, 0 < beta < 1. Generally close to 1.
beta_2: float, 0 < beta < 1. Generally close to 1.
epsilon: float >= 0. Fuzz factor. If None, defaults to K.epsilon().
decay: float >= 0. Learning rate decay over each update.
amsgrad: boolean. Whether to apply the AMSGrad variant of this algorithm from the paper "On the Convergence of Adam and Beyond".
```

hgaiser:
https://github.com/keras-team/keras/blob/master/keras/optimizers.py#L57

luliyucoordinate:
But when i run code , i find this `keras_retinanet/bin/train.py:101: UserWarning: Output "clipped_boxes" missing from loss dictionary. We assume this was done on purpose, and we will not be expecting any data to be passed to "clipped_boxes" during training.
  optimizer=keras.optimizers.adam()`

hgaiser:
Ah but I assume the warning stayed even after removing the clipnorm argument? The warning and the argument are unrelated.

Also: https://github.com/fizyr/keras-retinanet/blob/master/README.md#notes

hgaiser:
Closing this since I don't think we should delete this.

PR #380, opened 2018-04-10
add random augmentation to the training

Ori226:
I wanted to be able to apply extensive augmentation on the training images. What I suggest is a quick way to do it.

This is the first time I'm issuing a pull request in Github. Feel free to correct me if I'm doing something wrong.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -164,7 +164,21 @@ def create_callbacks(model, training_model, prediction_model, validation_generat
 
 def create_generators(args):
     # create random transform generator for augmenting training data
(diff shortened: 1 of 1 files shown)


de-vri-es:
Thanks for the PR. In general +1. The values look sane, and the old behaviour is preserved by default.

The only thing I wonder is if this is enough in terms of flexibility. Should we allow tweaking the values for each possible transform value individually or not? The obvious downside of that is that the user will also have to fill in much more values, whereas this gives a sane set of values.

We could also just go with this now, and perhaps allow more detailed fine-tuning with additional flags in the future if it comes up.

hgaiser:
Looks good to me too :+1: 

Successful first PR then :)

PR #385, opened 2018-04-12
Develop ori

Ori226:
I've added support for evaluating the other backend networks (mobilenet, vgg, densenet). I don't know why, but other commit from other users where also added. Please help me fix it (if it really shouldn't be here). I'm still  a git beginner :)


de-vri-es:
It looks like you did a rebase after a merge. The easiest way to solve this is to do a new rebase against the current master. If it doesn't remove the commits automatically, you can do an interactive rebase and drop the commits that you don't want to keep.

Or you can squash everything into a single commit. The PR is small enough that we'd probably do that anyway when merging.

PR #388, opened 2018-04-12
allow evaluation with backbone different from Resnet

Ori226:
Support evaluation of the other available back-ends. Please note that the back-ends are identify by the file name.

### DIFF
--- a/keras_retinanet/bin/evaluate.py
+++ b/keras_retinanet/bin/evaluate.py
@@ -34,7 +34,7 @@ from ..preprocessing.pascal_voc import PascalVocGenerator
 from ..preprocessing.csv_generator import CSVGenerator
 from ..utils.keras_version import check_keras_version
 from ..utils.eval import evaluate
-
+from ..models import custom_objects as custom_objects_importer
 
 def get_session():
     config = tf.ConfigProto()
@@ -113,18 +113,7 @@ def main(args=None):
 
     # create the generator
     generator = create_generator(args)
-
-    # load the model
-    if 'mobilenet' in args.backbone:
-        from ..models.mobilenet import custom_objects
-    elif 'resnet' in args.backbone:
-        from ..models.resnet import custom_objects
-    elif 'densenet' in args.backbone:
-        from ..models.densenet import custom_objects
-    elif 'vgg' in args.backbone:
-        from ..models.vgg import custom_objects
-    else:
-        raise NotImplementedError('Un-identified backbone. Not implemented.')
+    custom_objects = custom_objects_importer(args.backbone)
 
     # load the model
     print('Loading model, this may take a second...')
--- a/keras_retinanet/bin/evaluate.py
+++ b/keras_retinanet/bin/evaluate.py
@@ -34,7 +34,7 @@ from ..preprocessing.pascal_voc import PascalVocGenerator
 from ..preprocessing.csv_generator import CSVGenerator
 from ..utils.keras_version import check_keras_version
 from ..utils.eval import evaluate
-from ..models import custom_objects as custom_objects_importer
+from ..models import custom_objects
 
 
 def get_session():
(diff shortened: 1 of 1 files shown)


keras_retinanet/bin/evaluate.py
```diff
@@ -114,8 +112,22 @@ def main(args=None):
...
+    elif 'densenet' in os.path.basename(args.model):
+        from keras_retinanet.models.densenet import custom_objects
+    elif 'vgg' in os.path.basename(args.model):
+        from keras_retinanet.models.vgg import custom_objects
+    else:
+        raise NotImplementedError('Un-identified backbone. Not implemented.')
+
     # load the model
     print('Loading model, this may take a second...')
+
```
de-vri-es: Could you remove this blank line?

Ori226: yes

hgaiser:
Hmm the model name might not necessarily include the backbone. I'd be more in favour of adding an argument to define what backbone to use. Even better would be if we can find a way to encode in the model what backbone it contains, but I'm not sure how to do this.

rodrigo2019:
I'm working in a similar thing, here is a adapted example extracted from my code
I'm creating a callback that saves information like the losses value on a saved model 

```python
#write information
import h5py
f = h5py.File('../snapshots/tiny_darknet_pascal_104.h5','a')
f['Informations/backbone'] = 'tiny_darknet'
f.close()

#test model
import sys
sys.path.append("..")
from keras.models import load_model
from keras_retinanet.models.darknet import custom_objects

model = load_model('../snapshots/tiny_darknet_pascal_01.h5', custom_objects=custom_objects)

#print backbone information    
f = h5py.File('../snapshots/tiny_darknet_pascal_104.h5', 'r')
print(f['Informations/backbone'].value)
f.close()
```

de-vri-es:
One problem with that is that we'd need to know the path of the saved snapshots. While they are predictable, duplicating the logic to produce the file name is not very nice. It also requires re-opening the file to append data. I'm not sure how expensive that is with h5py.

Ideally we would tell keras to add some extra metadata, but save/load_model() doesn't currently seem to support that. Plus, load_model already needs to have the custom objects dictionary, so it can't use any metadata from the file to form the dictionary. That is solvable, but for now I think a command line flag is the best solution.

keras_retinanet/bin/evaluate.py
```diff
@@ -113,6 +113,7 @@ def main(args=None):
 
     # create the generator
     generator = create_generator(args)
+    custom_objects = custom_objects_importer(args.backbone)
```
hgaiser: Why as `custom_objects_importer`? It's more simple to import it as `custom_objects` and do : 
```python
model = keras.models.load_model(args.model, custom_objects=custom_objects(args.backbone))
```

keras_retinanet/bin/evaluate.py
```diff
@@ -89,6 +89,7 @@ def parse_args(args):
     parser.add_argument('--iou-threshold',   help='IoU Threshold to count for a positive detection (defaults to 0.5).', default=0.5, type=float)
     parser.add_argument('--max-detections',  help='Max Detections per image (defaults to 100).', default=100, type=int)
     parser.add_argument('--save-path',       help='Path for saving images with detections.')
+    parser.add_argument('--backbone', help='Backbone model used by retinanet.', default='resnet50', type=str)
```
hgaiser: `type=str` is the default, so can be omitted.

Also, please align on the `help` arg :)

keras_retinanet/bin/evaluate.py
```diff
@@ -116,7 +117,7 @@ def main(args=None):
 
     # load the model
     print('Loading model, this may take a second...')
-    model = keras.models.load_model(args.model, custom_objects=custom_objects)
+    model = keras.models.load_model(args.model, custom_objects=custom_objects(args.backbone))
```
hgaiser: I added a utility function for this: https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/models/__init__.py#L5

Ori226: @hgaiser  correct me if I'm wrong, but after the update you made, my PR is no longer needed. right?

de-vri-es: Err, it seems they did.

PR #392, opened 2018-04-13
Unexpected argument

lvaleriu:


hgaiser:
Ah damn, missed this one. Could you adjust the PR to pass the label to
class name function of the generator instead? (Same for the other PR)

lvaleriu:
Will do that today at some moment!

PR #395, opened 2018-04-15
add parse argument to change image sizes

rodrigo2019:


keras_retinanet/bin/train.py
```diff
@@ -355,6 +375,8 @@ def csv_list(string):
     parser.add_argument('--no-evaluation',   help='Disable per epoch evaluation.', dest='evaluation', action='store_false')
     parser.add_argument('--freeze-backbone', help='Freeze training of backbone layers.', action='store_true')
     parser.add_argument('--random-transform', help='Randomly transform image and annotations.', action='store_true')
+    parser.add_argument('--image-min-side', help='Freeze training of backbone layers.', type=int, default=800)
```
lvaleriu: Can you change the name of the help description?

rodrigo2019: For sure

rodrigo2019:
is there any reason for the resize_image function to give priority to image_max_side instead of image_min_side?
If you use images with high ratio and small size, this image can cause network error because it has a very small size

hgaiser:
LGTM!

> is there any reason for the resize_image function to give priority to image_max_side instead of image_min_side?
> If you use images with high ratio and small size, this image can cause error in the model, because it will has a very small side

I'm not sure actually. To be honest, the logic is the same as in `py-faster-rcnn`.

PR #408, opened 2018-04-18
Reduce complexity in FilterDetections.

hgaiser:
Was hoping this would reduce the memory footprint for inference, but it didn't.. nevertheless, the code is cleaner (less tensors to update).


de-vri-es:
Looks good to me, assuming it's tested.

PR #411, opened 2018-04-19
Optimize loss computations.

hgaiser:
Not sure if it really makes a difference w.r.t. speed, but this seems better.

Now it first filters out the values that it is going to ignore anyway, and then compute the loss on the remaining values.


de-vri-es:
Looks good :+1:, assuming it's tested

PR #417, opened 2018-04-20
Implement BackboneMeta class.

hgaiser:
This PR implements the `BackboneMeta` class, which holds meta information of backbones. This information can be used for backbone specific functions, such as imagenet weights downloading, custom objects, etc.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -80,10 +80,10 @@ def create_models(backbone_retinanet, backbone, num_classes, weights, multi_gpu=
     # optionally wrap in a parallel model
     if multi_gpu > 1:
         with tf.device('/cpu:0'):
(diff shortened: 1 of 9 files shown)


de-vri-es:
In general I think this is the right approach. I'm not very enthusiastic about the name `BackboneMeta` though. What about simply `Backbone` or if it has to be more complicated `BackboneDescription` or even `BackbonePlugin`?

hgaiser:
Would we then get a `ResNetBackbone` if we rename it to `Backbone`?

de-vri-es:
I think so.

hgaiser:
Updated, awaiting CI.

hgaiser:
Damn, nice catch :)

https://github.com/fizyr/keras-retinanet/commit/e4d8b9679b4b895b24b8eb889d6157bd50f1f523

de-vri-es:
cheer :beers:

PR #442, opened 2018-05-01
Use nearest method for upsampling.

hgaiser:
This PR uses the nearest interpolation method for upsampling (as described in the RetinaNet paper).


de-vri-es:
Why though? Nearest neighbour "interpolation" gives the worst possible image quality. Is that better for the network?

hgaiser:
Apparently so.

de-vri-es:
Since you tested and landed a slightly higher mAp, lgtm.

PR #449, opened 2018-05-03
Replace backbone with backbone_name in load_model example in README.md

iver56:
Ref https://github.com/fizyr/keras-retinanet/commit/afad72216a5d68c01ac9e6e487120656ce589950


hgaiser:
Thanks!

PR #463, opened 2018-05-16
Add documentation.

hgaiser:


keras_retinanet/backend/common.py
```diff
@@ -21,6 +21,18 @@
 
 
 def bbox_transform_inv(boxes, deltas, mean=None, std=None):
+    """ Applies deltas (usually regression results) to boxes (usually anchors).
```
de-vri-es: Maybe good to add the fact that it applies deltas and also undoes normalization.

keras_retinanet/utils/transform.py
```diff
@@ -14,13 +14,13 @@ def transform_aabb(transform, aabb):
     The result is a new AABB in the same coordinate system as the original AABB.
     The new AABB contains all corner points of the original AABB after applying the given transformation.
 
-    # Arguments
+    Args
         transform: The transformation to apply.
         x1:        The minimum X value of the AABB.
```
enricoliscio: Why is 'X' capital here?

hgaiser:
Forgot to fix your comments @enricoliscio , I fixed it on master now.

PR #464, opened 2018-05-16
Merge evaluate and evaluate_coco scripts.

hgaiser:


vcarpani:
Since its similar to the one we are using for maskrcnn, it's right for me

PR #467, opened 2018-05-19
Update CONTRIBUTORS.md from the git log.

de-vri-es:
A lot happened since the last time CONTRIBUTORS.md was updated.


hgaiser:
LGTM. Should we wait for people to possibly object being in the contributors file or shall we merge?

de-vri-es:
They're in the git log already. Plus I don't think many people will object to being credited for their contributions.

PR #468, opened 2018-05-22
Reordering callbacks to allow monitoring the mean average precision

apacha:
Two changes:
- Writing mAP into logs dictionary in `eval.py` to allow monitoring that value
- Reordering the creation of callbacks, so the `ModelCheckpoint` already sees the validation metric and can potentially use it, to only save the best model.


hgaiser:
Ah very nice, I didn't know `logs` worked like that. Could you do a similar thing for the COCO evaluation callback? It's [this](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/callbacks/coco.py#L30) metric that I think we should monitor.

hgaiser:
Thanks!

PR #472, opened 2018-05-24
Add flake8 to the testing

cclauss:
[__flake8__](http://flake8.pycqa.org) is a superset of pep8 so it runs all the normal pep8 tests plus other tests which find syntax errors and undefined names.

### DIFF
--- a/keras_retinanet/models/retinanet.py
+++ b/keras_retinanet/models/retinanet.py
@@ -169,6 +169,7 @@ class AnchorParameters:
     def num_anchors(self):
         return len(self.ratios) * len(self.scales)
 
+
 """
 The default anchor parameters.
 """
@@ -330,7 +331,7 @@ def retinanet_bbox(
         model = retinanet(num_anchors=anchor_parameters.num_anchors(), **kwargs)
 
     # compute the anchors
-    features = [model.get_layer(name).output for name in ['P3', 'P4', 'P5', 'P6', 'P7']]
+    features = [model.get_layer(p_name).output for p_name in ['P3', 'P4', 'P5', 'P6', 'P7']]
     anchors  = __build_anchors(anchor_parameters, features)
 
     # we expect the anchors, regression and classification values as first output
--- a/keras_retinanet/preprocessing/pascal_voc.py
+++ b/keras_retinanet/preprocessing/pascal_voc.py
@@ -132,10 +132,6 @@ class PascalVocGenerator(Generator):
         return truncated, difficult, box
 
     def __parse_annotations(self, xml_root):
-        size_node = _findNode(xml_root, 'size')
-        width     = _findNode(size_node, 'width',  'size.width',  parse=float)
-        height    = _findNode(size_node, 'height', 'size.height', parse=float)
-
         boxes = np.zeros((0, 5))
         for i, element in enumerate(xml_root.iter('object')):
             try:
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,4 +7,5 @@
 # E251 unexpected spaces around keyword / parameter equals
 # E501 line too long (85 > 79 characters)
 [tool:pytest]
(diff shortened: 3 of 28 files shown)


de-vri-es:
It looks like there are a number of useful diagnostics in there, so it seems like a good idea to switch to `flake8`. However, there are also some diagnostics that I consider unhelpful or even harmful for code clarity. In specific, I disagree with the diagnostics that disallow alignment of operators or `:` symbols .

Also, before we can merge this, the diagnostics actually need to be fixed. Otherwise the Travis tests won't pass.

Could you extend your PR by fixing the errors reported by flake8, either by following it's advice or in some cases by disabling the diagnostic?

cclauss:
Six remaining error types:
* __E305__ expected 2 blank lines after class or function definition, found 1
* __F401__ 'os' imported but unused
* __F403__ 'from .dynamic import *' used; unable to detect undefined names
* __F812__ list comprehension redefines 'name' from line 301
* __F821__ undefined name 'backbone'
* __F841__ local variable 'width' is assigned to but never used

tests/utils/test_transform.py
```diff
@@ -145,7 +145,7 @@ def test_transform_aabb():
 
 
 def test_change_transform_origin():
-    prng = np.random.RandomState(0)
+    np.random.RandomState(0)
```
de-vri-es: Looks like this can be removed entirely. This test doesn't need a PRNG.

keras_retinanet/preprocessing/pascal_voc.py
```diff
@@ -133,8 +133,8 @@ def __parse_annotation(self, element):
 
     def __parse_annotations(self, xml_root):
         size_node = _findNode(xml_root, 'size')
-        width     = _findNode(size_node, 'width',  'size.width',  parse=float)
-        height    = _findNode(size_node, 'height', 'size.height', parse=float)
+        width     = _findNode(size_node, 'width',  'size.width',  parse=float)  # noqa: F841
+        height    = _findNode(size_node, 'height', 'size.height', parse=float)  # noqa: F841
```
de-vri-es: Hmm, looks like we don't actually use the width and height here. Looks like we should just remove this then. @hgaiser: do you agree?

setup.cfg
```diff
@@ -8,3 +8,5 @@
 # E501 line too long (85 > 79 characters)
 [tool:pytest]
 pep8ignore = E201 E202 E203 E221 E241 E251 E402 E501
+flake8-max-line-length = 127
```
de-vri-es: :+1:

keras_retinanet/models/retinanet.py
```diff
@@ -330,7 +331,7 @@ def retinanet_bbox(
         model = retinanet(num_anchors=anchor_parameters.num_anchors(), **kwargs)
 
     # compute the anchors
-    features = [model.get_layer(name).output for name in ['P3', 'P4', 'P5', 'P6', 'P7']]
+    features = [model.get_layer(p_name).output for p_name in ['P3', 'P4', 'P5', 'P6', 'P7']]
```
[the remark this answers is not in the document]
de-vri-es: Ah I see, there is also an function argument called `name`. That explains.

cclauss:
Feel free to touch up and merge.

setup.cfg
```diff
@@ -8,3 +8,5 @@
...
+flake8-ignore = E201 E202 E203 E221 E241 E251 E402 E501
```
hgaiser: Doesn't this replace `pep8ignore` ?

keras_retinanet/preprocessing/pascal_voc.py
hgaiser: I agree :)

setup.cfg
cclauss: Yes..  Removed.

de-vri-es:
It's all looking good. Merging this. Thanks again for the PR :)

PR #478, opened 2018-05-29
Feature/allow removal of steps per epoch

mingruimingrui:
fit_generator currently requires steps_per_epoch as a compulsory input. However by adding __len__ to Generator, fit_generator automatically trains the model on all images in Generator object.

### DIFF
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -244,3 +244,25 @@ class Generator(object):
             self.group_index = (self.group_index + 1) % len(self.groups)
 
         return self.compute_input_output(group)
+
+    # When left unspecified, steps_per_epoch takes on the value of len(generator)
+    # https://keras.io/models/sequential/#fit_generator
+    # This allows the keras_retinanet.preprocessing.generator.Generator object to be callable by len
+    def __len__(self):
+        return self.size()
+
+    def __next__(self):
+        return self.next()
+
+    # DON'T IMPLEMENT __iter__
+    # Now we don't go all the way and turn this into an iterator and use this like
+    #
+    # for img, ann in <Generator object>:
+    #     ...
+    #     ...
+    #
+    # A generator is supposed to be callable by next infinitely while an iterator is not
+    # this function should return a copy of this object with an iterator of a fixed length
+    # to avoid complicated situations where fit_generator is unable to call next on this object
+    # def __iter__(self):
+    #     return self


keras_retinanet/preprocessing/generator.py
```diff
@@ -244,3 +244,25 @@ def next(self):
...
+    #
+    # for img, ann in <Generator object>:
+    #     ...
+    #     ...
+    #
+    # A generator is supposed to be callable by next infinitely while an iterator is not
+    # this function should return a copy of this object with an iterator of a fixed length
+    # to avoid complicated situations where fit_generator is unable to call next on this object
+    # def __iter__(self):
+    #     return self
```
de-vri-es: Could you remove the comment about not implementing `__iter__`?  If you want to preserve the reasoning, you can explain why we shouldn't implement `__iter__` in this PR or in the commit message.

mingruimingrui:
My bad, should have done something like don't merge this pull request or something.

`train_generator` should function perfectly as per before in `fit_generator`, with the added benefit that you don't have to define `steps_per_epoch`.

Now if you attempt to call `train_generator` as an iterator, it would throw a `NotImplementedError` since the ability that ability is disabled.

```
train_generator = CocoGenerator(...)

len(train_generator)
> return size of generator

next()
> return next input, target group

for inputs, targets in train_generator:
    ...
    ...
> will throw  NotImplementedError since the ability to call `train_generator` as iterator is disabled
```

keras_retinanet/preprocessing/generator.py
```diff
@@ -244,3 +244,25 @@ def next(self):
             self.group_index = (self.group_index + 1) % len(self.groups)
 
         return self.compute_input_output(group)
+
+    # When left unspecified, steps_per_epoch takes on the value of len(generator)
+    # https://keras.io/models/sequential/#fit_generator
+    # This allows the keras_retinanet.preprocessing.generator.Generator object to be callable by len
```
hgaiser: I like the addition of `__len__`, but the comment isn't required in my opinion.

keras_retinanet/preprocessing/generator.py
```diff
@@ -244,3 +244,25 @@ def next(self):
...
+    def __len__(self):
+        return self.size()
+
+    def __next__(self):
```
hgaiser: This function is already defined ?
https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/preprocessing/generator.py#L234-L235

keras_retinanet/preprocessing/generator.py
hgaiser: I agree with @de-vri-es

de-vri-es:
> Personally I am in favour of the latter, but many are in favour of the first. If you plan to add it to the training script, I would prefer to keep the default functionality as is and optionally add an argument --auto-steps or `--steps auto' or something to cover the other case.

I would prefer `--steps auto` even if that is probably slightly more difficult with pythons argparse.

de-vri-es:
> Now if you attempt to call train_generator as an iterator, it would throw a NotImplementedError since the ability that ability is disabled.

That's not new though, is it? Or do these changes disable that before somehow?

hgaiser:
What's the status of this PR? I see there are still some comments that are not addressed.

hgaiser:
Closing due to inactivity.

PR #479, opened 2018-05-29
Typo in variable name: 'backbone' --> 'self.backbone'

cclauss:
__backbone__ is an _undefined name_ in this context.

Discovered via https://travis-ci.org/fizyr/keras-retinanet/jobs/385214667#L751


de-vri-es:
Nice catch. Thanks!

PR #490, opened 2018-06-06
generator: Add Generator taking in input Pandas' Dataframe.

arthurdouillard:
I've used `keras-retinanet` for a project. It's an amazing library, congrats!

For my project I needed to use Pandas `DataFrame` instead of csv files, coco format, etc. Thus I wrote a generator that can take such input. I thought you may want it in the repository.

The in-code documentation is self-explanatory:

```python
class DfGenerator(CSVGenerator):
    """Custom generator intented to work with in-memory Pandas' dataframe."""
    def __init__(self, df, class_mapping, cols, base_dir='', **kwargs):
        """Initialization method.

        Arguments:
            df: Pandas DataFrame containing paths, labels, and bounding boxes.
            class_mapping: Dict mapping label_str to id_int.
            cols: Dict Mapping 'col_{filename/label/x1/y1/x2/y2} to corresponding df col.
        """
```

### DIFF
--- a/keras_retinanet/preprocessing/df_generator.py
+++ b/keras_retinanet/preprocessing/df_generator.py
@@ -1,0 +1,52 @@
+import numpy as np
+
+from keras_retinanet.preprocessing.csv_generator import CSVGenerator, Generator
+
+
+class DfGenerator(CSVGenerator):
+    """Custom generator intented to work with in-memory Pandas' dataframe."""
+    def __init__(self, df, class_mapping, cols, base_dir='', **kwargs):
+        """Initialization method.
+
+        Arguments:
+            df: Pandas DataFrame containing paths, labels, and bounding boxes.
+            class_mapping: Dict mapping label_str to id_int.
+            cols: Dict Mapping 'col_{filename/label/x1/y1/x2/y2} to corresponding df col.
+        """
+        self.base_dir = base_dir
+        self.cols = cols
+        self.classes = class_mapping
+        self.labels = {v: k for k, v in self.classes.items()}
+
+        self.image_data = self._read_data(df)
+        self.image_names = list(self.image_data.keys())
+
+        Generator.__init__(self, **kwargs)
+
+    def _read_classes(self, df):
+        return {row[0]: row[1] for _, row in df.iterrows()}
+
+    def __len__(self):
+        return len(self.image_names)
+
+    def _read_data(self, df):
+        data = {}
+        for _, row in df.iterrows():
+            img_file, class_name = row[self.cols['col_filename']], row[self.cols['col_label']]
+            x1, y1 = row[self.cols['col_x1']], row[self.cols['col_y1']]
+            x2, y2 = row[self.cols['col_x2']], row[self.cols['col_y2']]
+
+            if img_file not in data:
+                data[img_file] = []
+
+            # Image without annotations
+            if not isinstance(class_name, str) and np.isnan(class_name):
+                continue
+
+            data[img_file].append({
+                'x1': int(x1), 'x2': int(x2),
+                'y1': int(y1), 'y2': int(y2),
+                'class': class_name
+            })
+
+        return data


keras_retinanet/preprocessing/df_generator.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+
+from keras_retinanet.preprocessing.csv_generator import CSVGenerator, Generator
+
+
+class DfGenerator(CSVGenerator):
+    """Custom generator intented to work with in-memory Pandas' dataframe."""
```
hgaiser: intended*

keras_retinanet/preprocessing/df_generator.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        self.base_dir = base_dir
+        self.cols = cols
+        self.classes = class_mapping
+        self.labels = {v: k for k, v in self.classes.items()}
+
+        self.image_data = self._read_data(df)
+        self.image_names = list(self.image_data.keys())
+
+        Generator.__init__(self, **kwargs)
```
hgaiser: Prefer to use `super(CSVGenerator, self).__init__(self, **kwargs)` instead. Assuming your intention is to skip the CSVGenerator initializer and initialize the `Generator` instead.

keras_retinanet/preprocessing/df_generator.py
```diff
@@ -0,0 +1,52 @@
...
+
+class DfGenerator(CSVGenerator):
+    """Custom generator intented to work with in-memory Pandas' dataframe."""
+    def __init__(self, df, class_mapping, cols, base_dir='', **kwargs):
+        """Initialization method.
+
+        Arguments:
+            df: Pandas DataFrame containing paths, labels, and bounding boxes.
+            class_mapping: Dict mapping label_str to id_int.
+            cols: Dict Mapping 'col_{filename/label/x1/y1/x2/y2} to corresponding df col.
```
hgaiser: What does this do exactly? So the user has to input a dictionary here to map cols from their df to col_filename, col_label, etc. so that we can use those names in the generator?

Would it alternatively be okay to force the df to have cols named `filename, label, x1, y1, x2, y2` ? And if this is not the case, raise an exception.

de-vri-es:
> In general, is this a format that is quite normal for a Pandas df to have? Or is this a format designed by you which the module should enforce somehow?

In case it isn't clear, what @hgaiser wants to know is if the layout of the records is a (de-facto) standard way to store annotations in Pandas data frames or if it was specifically designed for `keras-retinanet`.

The reason we ask is that we're not entirely sure we want to have another generator for a custom format for `keras-retinanet`, but supporting a (de-facto) standard format could be useful.

keras_retinanet/preprocessing/df_generator.py
arthurdouillard: 1. Yes, so we don't force the user to name its Pandas df columns as we want.
2. If you prefer to have a fixed column names, sure I can do that.

arthurdouillard:
@de-vri-es @hgaiser Actually, if I understand you well, the layout of the records for Pandas is the same as a csv. It's just an overlay over a in-memory csv.

Without `DfGenerator`, I would have to serialize my Pandas dataframe to csv on disk(`df.to_csv('my_csv.csv', index=False`), then load it with your `CSVGenerator`. 

I'll make the asked changes this WE.

keras_retinanet/preprocessing/df_generator.py
hgaiser: I'm not sure. @de-vri-es what do you propose ?

de-vri-es: It could be a useful feature, but please drop the `col_` prefix from the keys. Since the argument is already called `cols` I think that's clear enough.

A default value which does an identity mapping probably make sense for the `cols` argument.

keras_retinanet/preprocessing/df_generator.py
de-vri-es: Why does this inherit from `CSVGenerator` at all? That seems unnecessary, and may prevent us from changing internal details of the `CSVGenerator` in the future.

de-vri-es:
> @de-vri-es @hgaiser Actually, if I understand you well, the layout of the records for Pandas is the same as a csv. It's just an overlay over a in-memory csv.

I suppose the real question is: why should we have a Pandas DataFrame backed generator? I'm not sure what the advantage is of having this. Does it ease integration with existing tools? Or is there some other reason why this improves the usability of `keras-retinanet`?

arthurdouillard:
It could be useful if one wants to load the csv in-memory, makes some train / val split in-memory (according to user arguments for example), then fed it to the network. It avoid to reserialize the csvs.

I found it useful for my use-case and wondered that you may want it. If you don't I understand.

hgaiser:
What is the status of this PR?

hgaiser:
I'm a bit conflicted on whether this is something we want or not. I believe that in case of doubt we should opt for not merging. Nonetheless, thank you for making the PR!

PR #495, opened 2018-06-08
Fix absent classes

vcarpani:


keras_retinanet/bin/evaluate.py
```diff
@@ -2,13 +2,10 @@
 
 """
 Copyright 2017-2018 Fizyr (https://fizyr.com)
-
```
hgaiser: Why are these removed?

keras_retinanet/bin/evaluate.py
```diff
@@ -132,23 +129,25 @@ def main(args=None):
...
-            score_threshold=args.score_threshold,
-            max_detections=args.max_detections,
-            save_path=args.save_path
-        )
-
-        # print evaluation
-        for label, average_precision in average_precisions.items():
-            print(generator.label_to_name(label), '{:.4f}'.format(average_precision))
-        print('mAP: {:.4f}'.format(sum(average_precisions.values()) / len(average_precisions)))
+    average_precisions = evaluate(
```
hgaiser: The way it is implemented now, `evaluate_coco` isn't used. COCO has quite a nice evaluation output, but it is unfortunately incompatible with other types of datasets. Still though, for COCO I think we should use `evaluate_coco`.

keras_retinanet/callbacks/eval.py
```diff
@@ -67,6 +76,4 @@ def on_epoch_end(self, epoch, logs=None):
         logs['mAP'] = self.mean_ap
 
         if self.verbose == 1:
-            for label, average_precision in average_precisions.items():
-                print(self.generator.label_to_name(label), '{:.4f}'.format(average_precision))
```
hgaiser: Printing the mAP per label can be quite useful. Why did you remove it?

keras_retinanet/bin/evaluate.py
carpo93: I have no idea why did it happened, will check

keras_retinanet/callbacks/eval.py
vcarpani: It is not removed, I just moved it inside the for loop at line 60

hgaiser: I see, then I haven't said anything :)

hgaiser:
It seems there are some unit test errors. If they're addressed, it LGTM.

PR #504, opened 2018-06-14
Keras 2.2.0

hgaiser:
These changes fix keras-retinanet for keras version 2.2.0. The only difference is that keras moved some imports to separate repositories and didn't preserve backwards compatibility..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
     - '2.7'
 install:
     - pip install 'numpy>=1.14'
-    - pip install 'keras==2.1.3'
+    - pip install 'keras==2.2.0'
     - pip install 'opencv-python>=3.3.0'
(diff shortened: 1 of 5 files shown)


carpo93:
Seems ok, but failed the test of python 3 while building test_train.py

de-vri-es:
With a segmentation fault no less. Looks like a problem with Travis or the PyPA releases of keras/tensorflow. Would be good to track down.

hgaiser:
Segfault appears to be introduced in keras  2.2.0. For more information, see [here](https://github.com/keras-team/keras/issues/10399). It appears calling `clear_session` or `set_session` multiple times within one process will segfault Keras. As a temporary fix I set the unit tests to run in a forked manner, so that every unit test runs in its own process. I'm awaiting the CI and then I'll merge.

Issue #466, opened 2018-05-18
How to deal with multiple predictions (with different classes) on a same target


netw0rkf10w:
Hello,

First of all thank you so much for the very valuable implementation! It works consistently well on diverse problems for me.

I would like to ask for your advice on an issue that I encountered often when using the code: a target can get two predicted boxes, with two different classes. How should I best deal with this?

Of course the simplest solution is to keep only the one with the highest score. I'm also thinking of another solution that is to take the average of the classification output vectors, then use that vector to predict the class.

Could I have your opinion please?
Thank you again!

hgaiser:
> I'm also thinking of another solution that is to take the average of the classification output vectors, then use that vector to predict the class.

I'm not sure I follow this approach? The classification vector is what is being used to perform the classification. If more than one of those classes have high scores, they are indeed two detection. This is by design actually, we didn't use to have that before (we'd only accept the maximum score for a classification vector, now we take any score above a threshold). I think the idea is that two objects can have very similar bounding boxes but still be separate classes.

I would guess that if your network is outputting two different classes for the same object, your objects are visually similar? Or maybe there are some mislabelled object in your dataset? Or maybe you don't have enough data for your use case?

There are many possibilities :)

netw0rkf10w:
@hgaiser Thanks for the answer!

> I'm not sure I follow this approach? The classification vector is what is being used to perform the classification. If more than one of those classes have high scores, they are indeed two detection. This is by design actually, we didn't use to have that before (we'd only accept the maximum score for a classification vector, now we take any score above a threshold).

If I understand correctly, you mean each bounding box can be assigned to two (or more) different classes? The problem I described above is a bit different though: I have two slightly different boxes (in terms of position), each is assigned to a different classes. Thus, I guess that these two boxes have two different classification vectors (because the boxes are not identical), so I can take the average of these vectors, hoping that will yield the true class (for my task I would like to have only one detection per object).

> I would guess that if your network is outputting two different classes for the same object, your objects are visually similar? Or maybe there are some mislabelled object in your dataset? Or maybe you don't have enough data for your use case?

Well all of these are likely, unfortunately!

My datset probably contains some noise, but I would say the level is not high. A more serious problem is the lack of diversity and balance: (a) the dataset contains only images of single class (i.e. each image contains only one class of objects, while the model will be used on images of different classes), and (b) the number of images can be very different between classes.

[...]

PR #512, opened 2018-06-19
Add option for class specific filtering.

hgaiser:
This PR adds a flag to disable class specific filtering. When disabled, each anchor would take the maximum scoring class and use that, instead of using all classes that have a score above a certain threshold.

### DIFF
--- a/keras_retinanet/bin/convert_model.py
+++ b/keras_retinanet/bin/convert_model.py
@@ -37,6 +37,7 @@ def parse_args(args):
     parser.add_argument('model_out', help='Path to save the converted model to.')
     parser.add_argument('--backbone', help='The backbone of the model to convert.', default='resnet50')
     parser.add_argument('--no-nms', help='Disables non maximum suppression.', dest='nms', action='store_false')
+    parser.add_argument('--no-class-specific-filter', help='Disables class specific filtering.', dest='class_specific_filter', action='store_false')
 
     return parser.parse_args(args)
 
--- a/keras_retinanet/layers/filter_detections.py
+++ b/keras_retinanet/layers/filter_detections.py
@@ -18,17 +18,27 @@ import keras
 from .. import backend
 
 
-def filter_detections(boxes, classification, other=[], nms=True, score_threshold=0.05, max_detections=300, nms_threshold=0.5):
+def filter_detections(
+    boxes,
+    classification,
+    other                 = [],
+    class_specific_filter = True,
+    nms                   = True,
+    score_threshold       = 0.05,
+    max_detections        = 300,
+    nms_threshold         = 0.5
+):
     """ Filter detections using the boxes and classification values.
 
     Args
-        boxes           : Tensor of shape (num_boxes, 4) containing the boxes in (x1, y1, x2, y2) format.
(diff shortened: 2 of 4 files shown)


netw0rkf10w:
I have tried this and it seems that NMS does not work properfly.

What I did is replacing the old code with the new one (downloaded from the [`class-specific-filtering` branch](https://github.com/fizyr/keras-retinanet/tree/class-specific-filtering)), reinstalling it (`pip install .`, because I'm using a virtual environment), then launching the training script from scratch. After 3 epochs I converted the model using 

`keras_retinanet/bin/convert_model.py --backbone resnet101 --no-class-specific-filter ./snapshots/resnet101_csv_03.h5 ./models/resnet101_csv_inference_03.h5`

and then used the converted model in my inference code.

Some results can be seen here: https://github.com/fizyr/keras-retinanet/issues/466#issuecomment-399096818

I have checked the values in `filter_detections.py` (two instances of them) and they are correct:

```
        nms                   = True,
        class_specific_filter = True,
        nms_threshold         = 0.5,
        score_threshold       = 0.05,
```

hgaiser:
So to be clear @netw0rkf10w , did  you use this branch, or did you copy parts of the code to master?

netw0rkf10w:
@hgaiser What I did is:
- Rename the folder containing the old code from `keras-retinanet` to `keras-retinanet_OLD`.
- Download the whole branch [class-specific-filtering](https://github.com/fizyr/keras-retinanet/tree/class-specific-filtering) and extract it to the same place as previous code: `keras-retinanet`.
- Go to `keras-retinanet` and `pip install .`.

Using the above steps I think I'm using the new code now.

hgaiser:
If you used `--no-class-specific-filter` argument, then yes you should be using the branch (otherwise it would say `--no-class-specific-filter` is an unknown argument).

netw0rkf10w:
@hgaiser Sorry I didn't know that you replied (I thought the notification was for the previous reply). Yes I used `--no-class-specific-filter` without errors.

netw0rkf10w:
@hgaiser To be sure I even trained my model from scratch with the new code.
Oh, I didn't know that you guys have an implementation of Mask R-CNN!!! That's awesome! I have been using matterport's one, but I'm going to test yours very soon! Thank you so much!

enricoliscio:
LGTM, you can merge

hgaiser:
> @hgaiser To be sure I even trained my model from scratch with the new code.
> Oh, I didn't know that you guys have an implementation of Mask R-CNN!!! That's awesome! I have been using matterport's one, but I'm going to test yours very soon! Thank you so much!

It's not advertised, but it's there and it's working :)

It's technically not really Mask-RCNN though. It's a combination of MaskRCNN and RetinaNet (see the disclaimer on keras-maskrcnn).

raeony:
@hgaiser @netw0rkf10w 
firstly, thanks for the great work you did!
I have tried the new codes you two discussed in my datasets, acctually the NMS didn't work still(the old codes didn't work yet). so I want to know what's  problem with this , do you have a solution to solve this?
the overlapping boxes are just like this:
image

netw0rkf10w:
@raeony Your results look strange. Which code are you using? Did you uninstall the previous version before installing the new one?

PR #520, opened 2018-06-21
Optimizations for data generator

hgaiser:
These optimizations should speed up the generator considerably. In my testing the time per batch increased from approximately 200msec to approximately 77msec.

PS. Apparently there are some unit test issues, I'll fix those tomorrow.


lvaleriu:
This is really nice indeed! Bravo!

hgaiser:
> This is really nice indeed! Bravo!

The downside is that installation now involves a (small) compilation step. I was hoping to avoid that.. but the cython version is simply much faster.

Also, I'm still fighting with Travis over the unit tests.

lvaleriu:
i already use to install cyton - it helps increasing debug speed in pycharm :). but more generally, increasing the generator speed justifies some effort. it will be a great idea maybe to provide a dockerfile for building an image "ready" for using your project (or at least some requirements installation script). this is something i do each time i spend some time on a project devellopment.

Issue #565, opened 2018-07-11
train.py throwing error while --val-annotations flag is enabled


sumeetssaurav:
File "keras_retinanet/bin/train.py", line 472, in <module>
    main()
  File "keras_retinanet/bin/train.py", line 467, in main
    callbacks=callbacks,
  File "/home/sumeet/.virtualenvs/virtual-py3/lib/python3.5/site-packages/keras/legacy/interfaces.py", line 91, in wrapper
    return func(*args, **kwargs)
  File "/home/sumeet/.virtualenvs/virtual-py3/lib/python3.5/site-packages/keras/engine/training.py", line 1426, in fit_generator
    initial_epoch=initial_epoch)
  File "/home/sumeet/.virtualenvs/virtual-py3/lib/python3.5/site-packages/keras/engine/training_generator.py", line 229, in fit_generator
    callbacks.on_epoch_end(epoch, epoch_logs)
  File "/home/sumeet/.virtualenvs/virtual-py3/lib/python3.5/site-packages/keras/callbacks.py", line 77, in on_epoch_end
    callback.on_epoch_end(epoch, logs)
  File "keras_retinanet/bin/../../keras_retinanet/callbacks/common.py", line 31, in on_epoch_end
    self.callback.on_epoch_end(epoch, logs=logs)
  File "keras_retinanet/bin/../../keras_retinanet/callbacks/eval.py", line 66, in on_epoch_end
    self.generator.label_to_name(label), 'with average precision: {:.4f}'.format(average_precision))
  File "keras_retinanet/bin/../../keras_retinanet/preprocessing/csv_generator.py", line 180, in label_to_name
    return self.labels[label]
KeyError: 0

hgaiser:
Did you by any chance start your class labels at 1? The evaluation tool doesn't like this much.

hgaiser:
Assuming this has been resolved.

PR #566, opened 2018-07-11
Ignore error for missing classes in CSVGenerator.

hgaiser:
Gaps can exist in the `CSVGenerator`, which results in errors when evaluating (https://github.com/fizyr/keras-retinanet/issues/565). This should ignore the KeyError when non-existing classes are used.

Alternatively (or in addition?) we could add a `has_label` and `has_classname` method to `Generator`, but I'm not sure if that is the correct approach.

### DIFF
--- a/keras_retinanet/preprocessing/csv_generator.py
+++ b/keras_retinanet/preprocessing/csv_generator.py
@@ -169,6 +169,16 @@ class CSVGenerator(Generator):
         """
         return max(self.classes.values()) + 1
 
+    def has_label(self, label):
+        """ Return True if label is a known label.
+        """
+        return label in self.labels
+
+    def has_name(self, name):
+        """ Returns True if name is a known class.
+        """
+        return name in self.classes
+
     def name_to_label(self, name):
         """ Map name to label.
         """
--- a/keras_retinanet/preprocessing/coco.py
+++ b/keras_retinanet/preprocessing/coco.py
@@ -75,6 +75,16 @@ class CocoGenerator(Generator):
         """
         return len(self.classes)
 
+    def has_label(self, label):
+        """ Return True if label is a known label.
+        """
+        return label in self.labels
+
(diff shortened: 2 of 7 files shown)


de-vri-es:
`label_to_name` isn't terrible when it doesn't work as it's mainly for visualization, but a missing `name_to_label` indicates an error in the dataset which should probably not be ignored silently.

Also, how about returning `None` instead of `'<unknown>'` for `label_to_name` for unknown labels?

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -169,14 +169,35 @@ def num_classes(self):
...
+        return name in self.classes
+
     def name_to_label(self, name):
         """ Map name to label.
+
+        Returns
+            The label of name. Returns -1 if it doesn't exist.
         """
+        if name not in self.classes:
+            raise ValueError("Invalid class name ({}) provided, options are {}.".format(name, self.classes.keys()))
```
de-vri-es: Should probably be a `KeyError`.

Also, it is probably better in python to just attempt the lookup and catch the `KeyError` and then throw a prettier key error.

keras_retinanet/preprocessing/csv_generator.py
```diff
@@ -169,14 +169,35 @@ def num_classes(self):
...
 
     def label_to_name(self, label):
         """ Map label to name.
+
+        Returns
+            The name of label. Returns '<unknown>' if it doesn't exist.
         """
+        if label not in self.labels:
+            return None
         return self.labels[label]
```
de-vri-es: This is basically `dict.get(key)`: https://docs.python.org/3/library/stdtypes.html#dict.get

`dict.get` may be more optimized though, since it's probably implemented in C,

hgaiser:
Updated again.

Issue #575, opened 2018-07-13
testing error: 'boxes, scores, labels ' not enough values to unpack (expected 3, got 2)


luvWY:
boxes, scores, labels = model.predict_on_batch(np.expand_dims(image, axis=0))

ValueError: not enough values to unpack (expected 3, got 2)

I dont why...can somebody solve it?

luvWY:
run "# Running directly from the repository:
keras_retinanet/bin/convert_model.py /path/to/training/model.h5 /path/to/save/inference/model.h5"

but the h5 file has error : load_model() got an unexpected keyword argument 'convert_model'
I dont know why

luvWY:
I fixed

hejiaget:
@luvWY I met the same probelm, could you please you how to solve it?

```
> boxes, scores, labels = model.predict_on_batch(np.expand_dims(image, axis=0))

ValueError: not enough values to unpack (expected 3, got 2)
```

luvWY:
convert your model into a inference model

Wulingtian:
@luvWY convert your model into a inference model,
TypeError: ('Keyword argument not understood:', u'class_specific_filter')

PhilippMarquardt:
The problem is that the parameter of the load_model function is called `convert `and not `convert_model`. Load the inference model like this (Just my example...you need to change the path obv): `model = models.load_model('snapshots/resnet50_pascal_03.h5', backbone_name='resnet50', convert=True)`

hgaiser:
Has this been resolved?

PhilippMarquardt:
I guess so.  If you take a look at the `load_model` function (https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/models/__init__.py) you can see that the parameter to convert the model is defined as `convert` and not `convert_model` like in the function call in the example notebook (https://github.com/fizyr/keras-retinanet/blob/master/examples/ResNet50RetinaNet.ipynb). 

Everyone who uses the example notebook will have this problem. I made a PR.

hgaiser:
Since the PR (#580) is merged, I will close this. Thanks @PhilippMarquardt !

PR #580, opened 2018-07-16
fix load_model call in the example notebook

PhilippMarquardt:
Just a quick fix of the `load_model` function call in the example notebook. See issue #575.


hgaiser:
Thanks!

PR #581, opened 2018-07-16
Add NCHW support.

hgaiser:
This PR adds support for NCHW order of images, which allegedly is faster when using CUDA/cuDNN than NHWC order.

### DIFF
--- a/keras_retinanet/backend/tensorflow_backend.py
+++ b/keras_retinanet/backend/tensorflow_backend.py
@@ -17,6 +17,12 @@ limitations under the License.
 import tensorflow
 
 
+def transpose(*args, **kwargs):
+    """ See https://www.tensorflow.org/versions/master/api_docs/python/tf/transpose .
+    """
+    return tensorflow.transpose(*args, **kwargs)
+
+
 def map_fn(*args, **kwargs):
     """ See https://www.tensorflow.org/versions/master/api_docs/python/tf/map_fn .
     """
--- a/keras_retinanet/layers/_misc.py
+++ b/keras_retinanet/layers/_misc.py
@@ -59,17 +59,24 @@ class Anchors(keras.layers.Layer):
 
     def call(self, inputs, **kwargs):
         features = inputs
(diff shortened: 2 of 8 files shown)


vcarpani:
Good for me, I trust you on the speed improvement, seems reasonable

hgaiser:
I don't know the exact values for the speedup, if there even is any. Regardless, it's good to support this. I'll try to measure the time difference later and update this PR with the results.

yhenon:
Have you check this converges properly?

In `retinanet.py`, you have this code:
```
    # reshape output and apply sigmoid
    outputs = keras.layers.Reshape((-1, num_classes), name='pyramid_classification_reshape')(outputs)
```

Which makes sense when using NHWC. But when changing to NCHW, the reshape is going to look different (basically n_classes will come from width, not channels). I think you might need a 
```
keras.layers.Permute((2,3,1))(outputs)
```

This also applies to regression.

hgaiser:
@yhenon very perceptive, I think you're right! Thanks for letting me know :)

hgaiser:
> I think u need to add (keras.layers.Permute((2,3,1))(f)) for i, f in enumerate(features) in build_anchors function

Hmm why do you think so? If I remember correctly, the anchors are a 3D tensor with shape `(batch_size, number_of_anchors, 4)`, meaning the `Permute` layer wouldn't work. Why do you think it has to be applied there?

ranti-iitg:
Anchor layer expects input to be NHWC, as    
'''
def call(self, inputs, **kwargs):

features = inputs
features_shape = keras.backend.shape(features)[:3]
'''
or we need to modify anchor layer.

hgaiser:
Ah, good catch! I prefer to change the anchor layer then, indeed. Thanks for the review :D

ranti-orbital:
looks grt

ranti-iitg:
looking great

vcarpani:
Tested with COCO, merging

PR #603, opened 2018-07-22
Tweak issue template.

de-vri-es:
This PR tweaks the issue template a bit.

Mainly, it changes the checkboxes into a numbered list and adds a request to remove the entire template to be replaced with the actual issue. The checkboxes sort-of imply that they should be ticked and remain in the issue, but I atleast prefer not to have them present in all issues.

I also changed the request for a script to these two points:
> * Clearly describe the issues you're having including the expected behaviour, the actual behaviour and the steps required to trigger the issue.
> * Include relevant output from the commands you're executing, including full stack traces where relevant.

I think in most situations it is more useful to know the commands being executed and their output than to always ask for a python script that triggers the issue. It can be a lot of work to make a minimal example, often for little gain. If something is complicated enough to warrant a minimum working example, we can still ask for one of course.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,13 +1,17 @@
-Please make sure that the boxes below are checked before you submit your issue. Only use GitHub issues for issues with the implementation, not for issues with specific datasets or general questions about functionality. If your issue is an implementation question, please ask your question on the [#keras-retinanet Slack channel](https://github.com/fizyr/keras-retinanet#discussions) and ask there instead of filing a GitHub issue.
+Please make sure that you follow the steps below when creating an issue.
+Only use GitHub issues for issues with the implementation, not for issues with specific datasets or general questions about functionality.
+If your issue is an implementation question, please ask your question on the #keras-retinanet Slack channel instead of filing a GitHub issue.
+You can find directions for the Slack channel here: https://github.com/fizyr/keras-retinanet#discussions
 
 Thank you!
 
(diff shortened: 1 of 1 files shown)


de-vri-es:
I also changed the formatting to be more readable when read as plain text, though at the cost of looking less nice when rendered. But people will mainly see it in plain text form when they're opening an issue.

ISSUE_TEMPLATE.md
```diff
@@ -1,13 +1,17 @@
-Please make sure that the boxes below are checked before you submit your issue. Only use GitHub issues for issues with the implementation, not for issues with specific datasets or general questions about functionality. If your issue is an implementation question, please ask your question on the [#keras-retinanet Slack channel](https://github.com/fizyr/keras-retinanet#discussions) and ask there instead of filing a GitHub issue.
+Please make sure that you follow the steps below when creating an issue.
+Only use GitHub issues for issues with the implementation, not for issues with specific datasets or general questions about functionality.
+If your issue is an implementation question, please ask your question on the #keras-retinanet Slack channel instead of filing a GitHub issue.
+You can find the Slack channel here: https://github.com/fizyr/keras-retinanet#discussions
```
hgaiser: Maybe change this to "You can find directions to the Slack channel here" ?

de-vri-es: Good point.

de-vri-es: Updated.

hgaiser: Then let's merge after Travis is done.

Issue #604, opened 2018-07-22
AttributeError: 'DictReader' object has no attribute 'next'


deep-diver:
File "./train.py", line 472, in <module>
    main()
  File "./train.py", line 420, in main
    train_generator, validation_generator = create_generators(args, backbone.preprocess_image)
  File "./train.py", line 284, in create_generators
    **common_args
  File "./../../keras_retinanet/preprocessing/open_images.py", line 202, in __init__
    self.annotations = generate_images_annotations_json(main_dir, metadata_dir, subset, cls_index, version=version)
  File "./../../keras_retinanet/preprocessing/open_images.py", line 79, in generate_images_annotations_json
    reader.next()
AttributeError: 'DictReader' object has no attribute 'next'

It didn't work for python 3.6.5 in my machine, so I changed to next(reader) instead.

relh:
deep-diver your fix works for me too so I made a PR.

hgaiser:
Closing this as #605 is merged.

PR #605, opened 2018-07-23
open_images.py quick fix

relh:
next(reader) replacing reader.next()

In Python 3.6.5 there is an AttributeError: 'DictReader' object has no attribute 'next' and this fixes it. #604


hgaiser:
Strange that it has no `next` method, but this fix seems harmless to me so if it fixes it for you, I'm find with merging this.

@lvaleriu ?

lvaleriu:
I'm using python 2.7 and the actual call "reader.next()" works. Ok for changing that to next(reader) (with a python version switch or direcly since this last call works in python 2.7 too)

hgaiser:
Seeing as it works on both python2 and python3, I don't believe a switch will be necessary. Thanks for the PR!

PR #621, opened 2018-08-04
debug.py --anchors option now shows the positive anchors

borakrc:
There was a bug that --anchors argument didn't show positive anchors in debug.py, but only the annotations. It is now fixed.

Old code screenshot:
image

New code screenshot:
image

Both taken with --anchors argument given to debug.py.


vcarpani:
Seems legitimate, what do you think @hgaiser ?

hgaiser:
Sorry for the late reply, but looks good to me, thanks!

PR #650, opened 2018-08-27
Fix iteration over dictionary

baek-jinoo:


de-vri-es:
Hey, thank you for your contribution!

One remark: `dict.iteritems()` is python2 only. In python3 the equivalent is simply 'dict.items()`. We mainly target python3, so could you swap it to `dict.items()`?

That is somewhat less efficient in python2 because it creates an actual list rather than an iterator. However, that's acceptable in this case.

baek-jinoo:
Thanks for the feedback. Made the changes. Please check when you get a chance.

de-vri-es:
Interesting, apparently this was never tested on python3. Thanks for changing the other iteritems/keys too!

PR #651, opened 2018-08-27
Add an interface to change anchors parameters

yecharlie:
Hi!
I write an anchors parameters interface. It gets the fllowing features:

- In the train.py, it supports an "--anchors" option, from which it load an anchors configure file in a human-readable yaml fromat (that is, a serialized dict object). By default, the model takes the default anchors parameters.

- Before it started to train the model, the anchors configure is saved beside the model if neccessary.  So that the next time one load the model to resumes the training process or start inferring process, he/she could omit the "--anchors" option. 

- The supported anchors parameters include ratios, scales, and also sizes, strides with the limitation that it should be in accordance with the Pyramid Feature Network architecture (ie: by default, the number of arguments in list ratios/scales is 5)
I have partially tested it on python3.6 (tests/bin/test_anchors.py).

Thank You!


enricoliscio:
Hi,

We are currently looking on this pull request. There will be updates soon. Thank you!

hgaiser:
Hey @yecharlie , thanks for your contribution. This work is being continued in https://github.com/fizyr/keras-retinanet/pull/651 , so I'll close this PR in favour of that one.

de-vri-es:
You mean of course #674

PR #658, opened 2018-08-30
Fix total mAP computation.

vcarpani:


enricoliscio:
LGTM. Add a little comment that it's the weighted sum over the total instances and not the classes, then you can merge.

hgaiser:
Maybe this should be an option and not the default? I agree this makes much more sense, but traditionally (COCO) it is not computed as a weighted sum.

vcarpani:
Sure, will fix it tomorrow, should we leave the COCO style as default?

hgaiser:
Yeah I think that makes sense.

enricoliscio:
LGTM. You can merge.

PR #683, opened 2018-09-14
Added PIL and cv2 as requirements in setup.py

adreo00:
Added PIL and cv2 as requirements in setup.py, as running train/eval/debug.py from clean virtual environment will result in error otherwise, even after running `pip install . --user`


vcarpani:
As discussed on slack, good idea, merging

PR #688, opened 2018-09-18
Added support for custom anchor parameters from config.ini in debug.py

adreo00:


keras_retinanet/bin/debug.py
```diff
@@ -177,7 +179,13 @@ def run(generator, args):
             image, image_scale = generator.resize_image(image)
             annotations[:, :4] *= image_scale
 
-        anchors = anchors_for_shape(image.shape)
+        # optionally load config parameters
+        anchor_params = None
```
enricoliscio: The config file parsing is written so to handle additional content in the future, hence it should not be uniquely linked to the anchor parameters. Please substitute these lines with:
```
   if args.config:
        args.config = read_config_file(args.config)

    anchor_params = None
    if args.config and 'anchor_parameters' in args.config:
        anchor_params = parse_anchor_parameters(args.config)
```

keras_retinanet/bin/debug.py
```diff
@@ -177,7 +179,13 @@ def run(generator, args):
...
+        if args.config:
+            config = read_config_file(args.config)
+            anchor_params = parse_anchor_parameters(config)
+
+        anchors = anchors_for_shape(image.shape,anchor_params=anchor_params)
```
enricoliscio: Please add a space between the two arguments of the function, after the comma, for styling and building purposes.

adreo00: Fixed

keras_retinanet/bin/debug.py
```diff
@@ -177,7 +184,12 @@ def run(generator, args):
             image, image_scale = generator.resize_image(image)
             annotations[:, :4] *= image_scale
 
-        anchors = anchors_for_shape(image.shape)
+        # optionally load anchor parameters
+        anchor_params = None
+        if args.config and 'anchor_parameters' in args.config:
+            anchor_params = parse_anchor_parameters(args.config)
```
adreo00: Would it make sense to move this block (line 187-190) outside of the run function into the main (to line 229)? anchor_params would also have to be passed to the run function, as in ` run(generator, args, anchor_params)`, but anchor_params shouldn't change from image to image and therefore doesn't need to be in the loop.

PR #700, opened 2018-09-26
Optional choose specific GPU to convert model

void32:
Option to select which GPU to use when converting a model as in _evaluate.py_.

### DIFF
--- a/keras_retinanet/bin/convert_model.py
+++ b/keras_retinanet/bin/convert_model.py
@@ -32,12 +32,14 @@ if __name__ == "__main__" and __package__ is None:
 # Change these to absolute imports if you copy this script outside the keras_retinanet package.
 from .. import models
 
+
 def get_session():
     """ Construct a modified tf session.
     """
     config = tf.ConfigProto()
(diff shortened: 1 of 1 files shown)


hgaiser:
Thanks for the PR! Actually I've thought about this before but never changed it, but could you change it to use the CPU instead of the GPU? There's no benefit to use a GPU here really.. Wrapping `models.load_model` with [this](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/bin/train.py#L107) should be sufficient (but please test it to make sure it is actually using the CPU :)).

void32:
Hi hgaiser,
Yes of course, I was also a bit surprised when i used the GPU:-) 
However, I tried wrapping with tf.device('/cpu:0') but it didn't work. Instead I had to set _CUDA_VISIBLE_DEVICES_ to empty.

hgaiser:
Alright, that'll work for me :)

Could you resolve the flake8 error? After that I'll merge it.

hgaiser:
Perfect, thanks!

PR #721, opened 2018-10-08
Refactor generators

hgaiser:
This PR refactors the generators so that the returned type of `load_annotations` is a dictionary. This allows for more customization in what data needs to be handled by the generator. Previously, all data would have to be transformed to fit together with the boxes / labels. With this PR, every piece is it's own key/value pair.

The minimum requirement for retinanet is to have `bboxes` and `labels` annotations.

### DIFF
--- a/keras_retinanet/preprocessing/coco.py
+++ b/keras_retinanet/preprocessing/coco.py
@@ -119,7 +119,7 @@ class CocoGenerator(Generator):
         """
         # get ground truth annotations
         annotations_ids = self.coco.getAnnIds(imgIds=self.image_ids[image_index], iscrowd=False)
-        annotations     = np.zeros((0, 5))
+        annotations     = {'labels': np.empty((0,)), 'bboxes': np.empty((0, 4))}
 
         # some images appear to miss annotations (like image with id 257034)
         if len(annotations_ids) == 0:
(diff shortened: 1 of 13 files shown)


keras_retinanet/preprocessing/coco.py
```diff
@@ -132,13 +132,12 @@ def load_annotations(self, image_index):
...
-        # transform from [x, y, w, h] to [x1, y1, x2, y2]
-        annotations[:, 2] = annotations[:, 0] + annotations[:, 2]
-        annotations[:, 3] = annotations[:, 1] + annotations[:, 3]
+            annotations['labels'] = np.concatenate([annotations['labels'], [self.coco_label_to_label(a['category_id'])]], axis=0)
+            annotations['bboxes'] = np.concatenate([annotations['bboxes'], [[
+                a['bbox'][0],
+                a['bbox'][1],
+                a['bbox'][0] + a['bbox'][2],
+                a['bbox'][1] + a['bbox'][3],
+            ]]], axis=0)
```
enricoliscio: Why are you making axis=0 explicit, and not in the csv_generator?

enricoliscio:
Looks good. Please, control why the checks are failing, and then you can merge.

PR #731, opened 2018-10-12
Fixed #712 by changing to non-zero initialization

### LINKED ISSUE #712: Initializing Conv Weights with Zero
### Where and What?

At the Code for the default_classification_model  ([retinanet.py#L66](https://github.com/fizyr/keras-retinanet/blob/078e12d87ecc9422f525965c0dc1a9478117c480/keras_retinanet/models/retinanet.py#L66)) the weights for the kernel are initialized with zero. According to the Keras Implementation (source code [here](https://github.com/keras-team/keras/blob/master/keras/initializers.py#L33)) it is actually 0.

This might improve model performance.

### Why is Zero Initialization a Big Deal?

As explained [here](https://stats.stackexchange.com/questions/200513/how-to-initialize-the-elements-of-the-filter-matrix):
"Why random? Why not initialize them all to 0? An important concept here is called symmetry breaking. If all the neurons have the same weights, they will produce the same outputs and we won't be learning different features. We won't learn different features because during the backpropagation step, all the weight updates will be exactly the same. So starting with a randomized distribution allows us to initialize the neurons to be different (with very high probability) and allows us to learn a rich and diverse feature hierarchy.

Why mean zero? A common practice in machine learning is to zero-center or normalize the input data, such that the raw input features (for image data these would be pixels) average to zero."

### Suggested Solution

Use a normal initializer such as:
```python
keras.initializers.normal(mean=0.0, stddev=0.01, seed=None)
```

So the whole Conv2D statement should be as follows:
```python
outputs = keras.layers.Conv2D(
        filters=num_classes * num_anchors,
        kernel_initializer=keras.initializers.normal(mean=0.0, stddev=0.01, seed=None),
        bias_initializer=initializers.PriorProbability(probability=prior_probability),
        name='pyramid_classification',
        **options
    )(outputs)
```

### Did I Test my Solution?

No, I do not have the resources to test it at the moment.

penguinmenac3:
As discussed this fixes the accidental zero initialization of a layer.

It very slightly improves APs on COCO.

### DIFF
--- a/keras_retinanet/models/retinanet.py
+++ b/keras_retinanet/models/retinanet.py
@@ -63,7 +63,7 @@ def default_classification_model(
 
     outputs = keras.layers.Conv2D(
         filters=num_classes * num_anchors,
-        kernel_initializer=keras.initializers.zeros(),
+        kernel_initializer=keras.initializers.normal(mean=0.0, stddev=0.01, seed=None),
         bias_initializer=initializers.PriorProbability(probability=prior_probability),
         name='pyramid_classification',
         **options


hgaiser:
Perfect, thanks!

PR #733, opened 2018-10-12
Assert outputs before converting to inference model.

hgaiser:
There have been quite some issues because people try to convert an already converted model. This PR adds a check that the outputs at least contain a `regression` and `classification` output (which isn't the case in a converted model).

### DIFF
--- a/examples/ResNet50RetinaNet.ipynb
+++ b/examples/ResNet50RetinaNet.ipynb
@@ -94,7 +94,7 @@
     "\n",
     "# if the model is not converted to an inference model, use the line below\n",
     "# see: https://github.com/fizyr/keras-retinanet#converting-a-training-model-to-inference-model\n",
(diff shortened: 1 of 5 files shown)


de-vri-es:
Good idea. Only the current implementation will result in a stack trace though if the assertion fails.

How about splitting this into a separate utility function to check if the model has the required outputs, so that at least `convert_model.py` can check it and print a clean error message without stack trace? That would be similar to the way the minimum keras version is checked in the scripts.

hgaiser:
I updated the PR, what do you think @de-vri-es ?

de-vri-es:
The extra function looks good, but I don't think the function `default_classification_model` should call `sys.exit` (indirectly). That's really something I would only do from the scripts themselves, but not from other library functions (apart from `check_training_model` itself).

enricoliscio:
Looks good, good idea.

PR #737, opened 2018-10-16
Add a --config option to the retinanet-convert-model command

ddowling:
If the model is trained with custom anchors read from a config.ini file then the conversion should also be run with the same configuration file to ensure the number of anchors is correct in the converted model.


enricoliscio:
Thank you for catching this, I had missed it. Everything looks good, but I don't see a reason why compiling the model after loading it. Why did you add it?

hgaiser:
Thanks for the PR! I wanted to release a new version of keras-retinanet today, so I made some changes to the PR to fix some conflict.

PR #746, opened 2018-10-19
Update ReduceLROnPlateau callback parameter name

lvaleriu:
keras warning: `epsilon` argument is deprecated and will be removed, use `min_delta` instead.


hgaiser:
Perfect, thanks !

PR #766, opened 2018-10-27
Improvement of workflow to prepare your own dataset in CSV format

akchan:
# Summary
This pull request will add an introduction of a sample workflow to prepare own dataset in CSV.

# Problem
Thank your for your great effort in `keras-retinanet`.

The CSV format included in `keras-retinanet` is very nice way to prepare own dataset for training. I think many people use a tool to annotate images, but couldn't find a tool that can create an annotation file in CSV format used in `keras-retinanet` directly.

# Solution
So, I create a format converting script `conv_pascal_to_csv.rb` and pull request here to introduce this workflow.

# Discussion point of this pull-req
Modification of README.md in this pull-req will introduce a specific tool. I'm worried if someone thinks that this pull-req isn't appropriate in this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,16 @@ cat,1
 bird,2
 ```
 
+
+### Convert annotation format to CSV
+Here is a sample workflow to prepare your own datase.
+
+1. Correct images for training
+2. Create annotations of images using [LabelImg](https://github.com/tzutalin/labelImg) with Pascal VOC xml format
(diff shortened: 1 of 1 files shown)


hgaiser:
Hey, sorry for getting back to this PR so late.

I don't think we should merge this, the steps listed to make a csv dataset are complicated, while the csv format is intended to be simple and straightforward. Thank you for the contribution though.

PR #786, opened 2018-11-07
Float16 fixes

vcarpani:
This PR enables running with FP16.
Waiting on this PR on keras:
https://github.com/keras-team/keras/pull/11580

### DIFF
--- a/keras_retinanet/preprocessing/generator.py
+++ b/keras_retinanet/preprocessing/generator.py
@@ -33,7 +33,7 @@ from ..utils.image import (
     apply_transform,
     preprocess_image,
     resize_image,
-    postprocess_image,
+    cast_image_to_floatx,
 )
 from ..utils.transform import transform_aabb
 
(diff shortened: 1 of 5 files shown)


keras_retinanet/utils/image.py
```diff
@@ -60,6 +63,20 @@ def preprocess_image(x, mode='caffe'):
     return x
 
 
+def cast_image_to_floatx(x):
```
hgaiser: `keras.backend.cast_to_floatx` also works on np arrays

keras_retinanet/preprocessing/generator.py
```diff
@@ -53,6 +54,7 @@ def __init__(
         compute_anchor_targets=anchor_targets_bbox,
         compute_shapes=guess_shapes,
         preprocess_image=preprocess_image,
+        cast_image_to_floatx=cast_image_to_floatx,
```
hgaiser: I don't think we need to pass this to the generator as a functor.

vcarpani:
@hgaiser the PR on keras is merged

keras_retinanet/utils/image.py
hgaiser: I think we can remove this wrapper entirely.

hgaiser:
I have one comment, but after that it can be merged as far as I'm concerned.

PR #791, opened 2018-11-10
Small questioning ...

lvaleriu:
Wouldnt it be better to print both mAP values since the evaluation takes some time already for some datasets? It will take twice this time if we want to compute these values separately... (this is happening too me right now and I really start to find this annoying :) ).


de-vri-es:
Merging.

PR #795, opened 2018-11-13
Avoid importing numpy in setup.py

maxclaey:
With the current setup.py file, keras-retinanet cannot be installed if numpy is not present.
With the updated version, numpy is specified as a dependency of keras-retinanet and is installed on the fly if needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,38 @@
 import setuptools
 from setuptools.extension import Extension
+from distutils.command.build_ext import build_ext as DistUtilsBuildExt
 
-class custom_wrapper(object):
-    def __new__(cls, *args, **kwargs):
-        from setuptools.command.build_ext import build_ext
-        class custom_build_ext(build_ext, object):
-            def run(self):
-                import numpy as np
-                self.include_dirs.append(np.get_include())
-                build_ext.run(self)
-        return custom_build_ext(*args, **kwargs)
+
+class BuildExtension(setuptools.Command):
+    description     = DistUtilsBuildExt.description
+    user_options    = DistUtilsBuildExt.user_options
+    boolean_options = DistUtilsBuildExt.boolean_options
+    help_options    = DistUtilsBuildExt.help_options
+
+    def __init__(self, *args, **kwargs):
+        from setuptools.command.build_ext import build_ext as SetupToolsBuildExt
+
+        # Bypass __setatrr__ to avoid infinite recursion.
+        self.__dict__['_command'] = SetupToolsBuildExt(*args, **kwargs)
+
+    def __getattr__(self, name):
+        return getattr(self._command, name)
+
+    def __setattr__(self, name, value):
+        setattr(self._command, name, value)
+
+    def initialize_options(self, *args, **kwargs):
+        return self._command.initialize_options(*args, **kwargs)
+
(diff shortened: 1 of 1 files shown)


de-vri-es:
+1 for not required numpy in `setup.py`. However, why do we need it in the first place? Can't we just declare numpy as a dependency and be done with it?

maxclaey:
I assume that to build the compute_overlap extension, Cython has to know where to find the numpy header files, hence the np.get_include(). I tried to solve this by adding numpy as a setup dependency, and only importing numpy at runtime. However, this would just move the problem to requiring Cython. Therefore, the custom_wrapper closure was needed. Now all works fine when installing with pip, but unfortunately, building with setup.py fails (causing the failed Travis CI build). At the moment I don't see an easy way to solve this without importing Cython earlier, and thus causing the same problem as we initially had with numpy.

de-vri-es:
hmm, I see.. I would really like if we can install properly with pip and `setup.py`. Any idea why running `setup.py` directly doesn't work anymore?

maxclaey:
The error that pops up is the following:
`distutils.errors.DistutilsClassError: command class <class '__main__.custom_wrapper'> must subclass Command`
The problem is that custom_wrapper should extend the Cython build_ext, but we can't do that immediately, as that would require us to import Cython. That's why the closure is used, and constructing custom_wrapper returns a custom_build_ext that dóes extend build_ext. But apparently some checks are done on custom_wrapper itself. I'm by far no expert on this, so I don't really know how to circumvent this.

de-vri-es:
I pushed a commit which I think should fix the issue.

I'm not sure why you had to create an additional wrapper class, but this seems to work as intended. To test, I removed both cython and numpy from my system, then ran `pip install --user .`.  It downloaded all the required dependencies (including numpy and cython) without causing any import errors.

Running `python setup.py build_ext --inplace` directly also worked.

@maxclaey: Could you review the commit I pushed?

maxclaey:
I just tested it again in a new fresh environment, just to be sure that I didn't overlook anything yesterday, but I get the same problem again. This is part of the output of `pip install .`
```running build_ext
  building 'keras_retinanet.utils.compute_overlap' extension
  error: unknown file type '.pyx' (from 'keras_retinanet/utils/compute_overlap.pyx')
  
  ----------------------------------------
  Failed building wheel for keras-retinanet```
This is because because when you import build_ext from setuptools.command.build_ext, it checks if Cython is present. It needs Cython to be able to handle .pyx files. That's why I needed the extra wrapper to be sure to only load build_ext when Cython is already installed. It's an option to provide a precompiled compute_overlap.c file as a fallback for when Cython is not present, but I'm not sure that's the way to go..

de-vri-es:
I don't understand why I don't get that error. I can run `pip install --user .` fine without cython installed. Which version of python and pip are you using exactly?

I'm on python 3.7.1 and pip 18.0

de-vri-es:
Hmm, weird. However, I pushed another attempt that still works for me with both pip and `setup.py` invoked directly.

maxclaey:
Just tested your new approach, works like a charm this time. Also tests are passing. Great work @de-vri-es , thanks a lot!

de-vri-es:
Good to hear, thanks for the PR, and thanks for testing. Merging :)

PR #841, opened 2018-12-16
Fix #840

### LINKED ISSUE #840: Broken `--workers` option handling
On python 3.6.3 and latest `keras-retinanet` from master if I specify `--workers` I get the following error:
```
Traceback (most recent call last):
  File "/usr/local/bin/retinanet-train", line 11, in <module>
    load_entry_point('keras-retinanet', 'console_scripts', 'retinanet-train')()
  File "/root/data/gishack/keras-retinanet/keras_retinanet/bin/train.py", line 487, in main
    if args.workers > 0:
TypeError: '>' not supported between instances of 'str' and 'int'
```

ei-grad:


ei-grad:
Oops, found that there is a second option :).

PR #856, opened 2019-01-02
Fix loading weights even if --no-weights is set.

### LINKED ISSUE #588: [bug report]Using VGG19 as backbone
Hi,

I want to use VGG19 as backbone. But I encountered the issue "AttributeError: module 'keras.applications.vgg19' has no attribute 'WEIGHTS_PATH_NO_TOP'.

Any ideas? Thanks a lot

martinzlocha:
When VGG was created it would use the imagenet weights even if --no-weights was set. This should be now fixed.

It should also fully fix #588.


hgaiser:
LGTM, thanks! Will merge when tests have passed.

PR #824, opened 2018-12-05
keras sequence methods for generator

bw4sz:
The goal of this PR is to introduce the keras sequence method (https://keras.io/models/sequential/) to the default generator class.

Goals 

1. Improve parallelization and thread-safe generators. Keras docs state:

"The use of keras.utils.Sequence guarantees the ordering and guarantees the single use of every input per epoch when using use_multiprocessing=True."

2. Implemented a __getitem__ and __len__ method.

3. Remove threading and locks.

4. Move group shuffling to explicit on_epoch_end method

5. Added command line arguments for the fit_generator method.

I cloned the master, built a new conda env, and ran the generator and generator csv tests. 

*Potential Drawbacks*

A number of windows users have complained about multiprocessing in keras. I found that multiprocessing was only useful when

1. Generator performs alot of processing. This could happen when files are very large and reading is the bottleneck. Or when the user creates a custom generator method (as I've done in my fork).

2. When there is sufficient memory to fill the queue. 

My production environment:

I am running linux on a large multi-node cluster, feeding in > 20,000 images during training in batches of 6. I found a huge performance boosts from using the sequence method and multiprocessing, but the overhead is probably too much for the csv generator method in most cases.

PR #873, opened 2019-01-16
Fixed shuffling at end of epoch when shuffling is False

LaurensHagendoorn:
Fixed this small bug that was introduced with https://github.com/fizyr/keras-retinanet/pull/824
It's currently shuffling every epoch, with this patch, it takes the shuffle_groups parameter into account again.


de-vri-es:
Looks good. Thank you for noticing and fixing the bug! :)

PR #884, opened 2019-01-29
Add validation data to training, and return Keras History object.

dshahrokhian:
The reason for this addition is to take into account the validation generator during training (for validation loss, etc.), and return the History object from Keras for logging purposes instead of just ignoring it. 

Please have a look at [History Callback](https://keras.io/callbacks/#history) for more information on the usefulness of this addition.

Best,
Daniyal

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -493,7 +493,7 @@ def main(args=None):
         use_multiprocessing = False
 
     # start training
-    training_model.fit_generator(
+    return training_model.fit_generator(
         generator=train_generator,
         steps_per_epoch=args.steps,
         epochs=args.epochs,
@@ -501,7 +501,8 @@ def main(args=None):
         callbacks=callbacks,
         workers=args.workers,
         use_multiprocessing=use_multiprocessing,
-        max_queue_size=args.max_queue_size
+        max_queue_size=args.max_queue_size,
+	validation_data=validation_generator
     )
 
 
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -502,7 +502,7 @@ def main(args=None):
         workers=args.workers,
         use_multiprocessing=use_multiprocessing,
         max_queue_size=args.max_queue_size,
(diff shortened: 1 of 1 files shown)


hgaiser:
As far as I interpret this PR, it should compute validation losses and return the `History` object from `main`. Why should we return the `History` object? It is not being used in `train.py` at least. Are you importing `train.py` and running `main()`, and doing something with the `History` object?

dshahrokhian:
> Why should we return the History object? It is not being used in train.py at least. Are you importing train.py and running main(), and doing something with the History object?

Yes exactly. That is the simplest way to use this repo as a dependency, without needing to duplicate `train.py`.

hgaiser:
Okay, fair enough. Could you fix the style error reported by Travis? Also, I think we should make computing the validation loss optional. It means processing all (validation) data so we shouldn't do it without explicitly requiring it. The default should disable the validation loss.

dshahrokhian:
Sure, will do. 

The validation dataset on its own is optional (see https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/bin/train.py#L387), and if the users are providing a validation set it is because they are interested into validation metrics, and to compare the training/validation loss curves to determine convergence, overfitting, etc.

Are you sure it makes sense to add yet another optional argument?

dshahrokhian:
True, but there is no way around it as both metrics are independent. I can add the extra parameter, but would you kindly explain why the build is not passing? `AttributeError: 'Application' object has no attribute 'make_notifier'`

hgaiser:
> would you kindly explain why the build is not passing? `AttributeError: 'Application' object has no attribute 'make_notifier'`

Good question, no clue :p 

Do you get the same error if you run the flake8 tests locally ?

`pytest --flake8`

hgaiser:
Yay, rerun the tests and it magically passes now. Could you implement the argument for enabling computing the validation loss?

dshahrokhian:
done!

tonmoyborah:
Has this been merged? It is much needed.

hgaiser:
It is now :)

Issue #781, opened 2018-11-05
issue with debug.py when input csv contains images with no objects of interest


amruz:
I am checking my dataset with debug.py, but its throwing an error when my csv file has no objects of interest as below.

> /keras-retinanet/objectdetection_obj5/falseimages/resized12.jpg,,,,,

The error is as follows

> Traceback (most recent call last):
  File "debug.py", line 242, in <module>
    main()
  File "debug.py", line 238, in main
    run(generator, args, anchor_params=anchor_params)
  File "debug.py", line 189, in run
    positive_indices, _, max_indices = compute_gt_annotations(anchors, annotations['bboxes'])
  File "../../keras_retinanet/utils/anchors.py", line 140, in compute_gt_annotations
    argmax_overlaps_inds = np.argmax(overlaps, axis=1)
  File "/volume/software/common/foreign_packages/NumPy/1.14.5/osl42-x86_64/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 1004, in argmax
    return _wrapfunc(a, 'argmax', axis=axis, out=out)
  File "/volume/software/common/foreign_packages/NumPy/1.14.5/osl42-x86_64/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 52, in _wrapfunc
    return getattr(obj, method)(*args, **kwds)
ValueError: attempt to get argmax of an empty sequence

Any guidance is highly appreciated
Thank you in advance

hgaiser:
This could be an issue with debug.py, but I have no time to look at it currently. Could you investigate it?

hgaiser:
Any update on this ?

Etienne-Meunier:
Hey I got the same error, and the thing is that we send an empty annotation dictionnary to compute_gt_annotations so it cannot extract the indices. 
I don't think it's the best way to do but if you place  a condition : if len(annotations['labels']) != 0 
around line 178 to line 202 it will just display the image for thoses cases and so no issue will happen

hgaiser:
A PR to solve this issue would be nice.

Etienne-Meunier:
I just did a Pull request with the change I mentioned upper 😄

hgaiser:
The PR is merged, so I'm closing this issue. Thanks!

PR #913, opened 2019-02-19
FIX to - issue #781

Etienne-Meunier:
Issue with debug.py when input csv contains images with no objects of interest #781

Just added a condition to avoid trying to avoid checking annotations where there is no object of interest.

### DIFF
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -175,31 +175,31 @@ def run(generator, args, anchor_params):
         # load the data
         image       = generator.load_image(i)
         annotations = generator.load_annotations(i)
+        if annotation['labels'] :
+            # apply random transformations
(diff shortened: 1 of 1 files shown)


hgaiser:
Thanks for the contribution! To avoid a lot of indentations, could you use the inverse of your check? This would be something like:

```python
if len(annotations['labels']) == 0 :
    continue
```

Etienne-Meunier:
Yep, of course, but we have to run the cv2.imshow after so it will something like this : 
if len(annotations['labels']) == 0 :
    cv2.imshow('Image', image)
    continue

otherwise the image will not be displayed where there is no annotation 
is it better ?

hgaiser:
Oh right, missed that part! Nah code duplication is worse, let's keep it as is :) thanks!

PR #945, opened 2019-03-17
Anchors grow dynamically with smaller strides

gergopool:
I'd like to detect smaller objects, so I made a bit of research on how the anchor parameter settings work in your repository. Since small objects might appear almost anywhere in the image, I'd like to place the anchorboxes tight and decrease the default anchorbox.strides parameter (from [8, 16, 32, 64, 128]). 

Looking at shift() and guess_shapes() functions under utils.anchorbox module, when I changed the default strides to [4, 8, 16, 32, 64] (halving the default), the anchorboxes covered only the top-left quarter of the image. This tiny change ensures the anchors cover the entire image, not less and not more, whatever strides are given.

Sorry for making it in 5 commits, I messed things up. ^^'

### DIFF
--- a/keras_retinanet/utils/anchors.py
+++ b/keras_retinanet/utils/anchors.py
@@ -188,7 +188,7 @@ def guess_shapes(image_shape, strides):
 
     Args
          image_shape: The shape of the image.
-         pyramid_levels: A list of what strides are used.
+         strides: A list of what strides are used.
 
     Returns
         A list of image shapes at each pyramid level based on the stride size
--- a/keras_retinanet/utils/anchors.py
+++ b/keras_retinanet/utils/anchors.py
@@ -183,7 +183,7 @@ def make_shapes_callback(model):
     return get_shapes
 
 
-def guess_shapes(image_shape, pyramid_levels):
+def guess_shapes(image_shape, strides):
     """Guess shapes based on pyramid levels.
 
     Args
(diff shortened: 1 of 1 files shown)


hgaiser:
Hmm did this work? I always thought that the strides and sizes were tied to how you construct your feature pyramid network. If you change one you should change the other accordingly.

I'll think about it some more, but I would be interested to hear what your experience is.

gergopool:
> Hmm did this work? I always thought that the strides and sizes were tied to how you construct your feature pyramid network. If you change one you should change the other accordingly.
> 
> I'll think about it some more, but I would be interested to hear what your experience is.

Take a few images & bounding boxes, divide all strides with 4 and check if the ground truth boxes are covered in your debug tool. I made my shot and all ground truth boxes turned red except the ones in the top left corner.

hgaiser:
I meant did the network, after training, correctly detected objects? The FPN downscales the images in a certain way, which matches the default strides and sizes. I wouldn't be surprised that if you change the strides and sizes then the network fails to detect objects properly.

hgaiser:
@gergopool any update on my previous question ?

gergopool:
> @gergopool any update on my previous question ?

That project of mine got left behind, so I didn't check who's right.
I think you're right and the strides have a strong relationship with the backbone, because if we reduce the size of the image by 8, each pixel on the feature map should represent a 8x8 grid on the original image. On the other hand, you provided the strides as optional parameters, but actually it looks that should never be changed.

With my changes I can only ensure that the debug.py shows the right anchor boxes if the 'strides' parameter is changed, but didn't train the model afterwards - so it might be completely rubbish.

hgaiser:
Ah I see. Yeah theoretically you need to change it when you change the backbone, but practically that never occurs. I will close this PR for now, but if you feel there is new evidence or something to discuss, feel free to open.

pleaseRedo:
@hgaiser 
I recently changed the backbone from resnet50 to DetNet(self-implement). C4 and C5 in DetNet does not downsample themself(stride = 1) as they are using dilated-conv. 
I think in this case we do need to consider changing strides for anchors right?

hgaiser:
I would assume so, yes.

Issue #976, opened 2019-04-17
val_loss not available (validation set and --compute-val-loss are provided)


mrtgenet:
Hi,

**First thanks for the helpful tools you provide here ! Really appreciate :)**

Il searched a bit through open and closed issues but I didn't manage to find answer, though some issues are related. So let me quickly explain : 

I am training a net with train and val sets, all works fine with default settings and I have promising results. But I would like to check validation loss instead of training loss (saving only best snapshots), which should be possible by changing the monitor value to 'val_loss' in the ModelCheckPoint callback and providing the "--compute-val-loss" argument. 

But when I tried, I got the following error : 
**"RuntimeWarning: Can save best model only with val_loss available, skipping.
  'skipping.' % (self.monitor), RuntimeWarning)"**

Here the corresponding code section in train.py : 
```
# save the model
if args.snapshots:
  # ensure directory created first; otherwise h5py will error after epoch.
  makedirs(args.snapshot_path)
  checkpoint = keras.callbacks.ModelCheckpoint(
    os.path.join(
      args.snapshot_path,
      '{backbone}_{dataset_type}_{{epoch:02d}}.h5'.format(backbone=args.backbone, dataset_type=args.dataset_type)
    ),
    verbose=1,
    save_best_only=True,
    monitor="val_loss",
    mode='min'
)
checkpoint = RedirectModel(checkpoint, model)
callbacks.append(checkpoint)
```

And here is the command I run  : 
```
keras_retinanet/bin/train.py
  --weights snapshots/ResNet-50-model.keras.h5 
  --batch-size 12 
  --epochs 90 
  --steps 1000 
  --freeze-backbone 
  --random-transform 
  --config keras_retinanet/config.ini 
  --compute-val-loss 
  csv path/to/train_set.csv path/to/class_map.csv --val-annotations path/to/val_set.csv
```

Is there something wrong or do I misunderstand the "--compute-val-loss" argument ?

Also I checked the code and I found that "--compute-val-loss" action in arg parser is 'store_false', is it normal ? It seems weird at first glance, though I must tell I didn't looked through the whole code to see how exactly it is used.

Thank you for any tip, 
Martin

liminghuiv:
Hi @mrtgenet ,
I had similar issue before. '--compute-val-loss' is not available. I would suggest that we add it, it is useful.

de-vri-es:
[...]

PR #978, opened 2019-04-18
compute_val_loss parser action to store_true

mrtgenet:
Either the argument should be named "--no-val-loss" or something like this and have a store_false action, or be named "--compute-val-loss" and have a store_true action. Second option is more coherent with the fact that providing a validation set is optional.

See [this issue](https://github.com/fizyr/keras-retinanet/issues/976)


de-vri-es:
Looks good. Once CI passes, we can merge. Thanks for the PR!

de-vri-es:
It passed! \o/

PR #997, opened 2019-05-09
Check validity of the config to avoid silent errors.

martinzlocha:
The configparser library which is used fails silently when it cannot find the config file and then retinanet continues as normal and uses the default configuration. This PR adds assertions to check that the file exists and isn't malformed.

### DIFF
--- a/keras_retinanet/utils/config.py
+++ b/keras_retinanet/utils/config.py
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-import os
 import configparser
 import numpy as np
 import keras
--- a/keras_retinanet/utils/config.py
+++ b/keras_retinanet/utils/config.py
@@ -23,15 +23,18 @@ from ..utils.anchors import AnchorParameters
 
 def read_config_file(config_path):
     config = configparser.ConfigParser()
-    config.read(config_path)
 
-    assert os.path.isfile(config_path), "Could not find {}.".format(config_path)
+    with open(config_path, 'r') as file:
+        config.read_file(file)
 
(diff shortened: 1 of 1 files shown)


keras_retinanet/utils/config.py
```diff
@@ -24,6 +25,14 @@ def read_config_file(config_path):
     config = configparser.ConfigParser()
     config.read(config_path)
 
+    assert os.path.isfile(config_path), "Could not find {}.".format(config_path)
```
de-vri-es: This check is now after attempting to read the file. I would much prefer if you switch to something like this:
```python
with open(config_path, 'r') as file:
    config.read_file(file)
```

That will throw a regular `FileNotFoundError`, which makes it easier to catch the error if desired.

martinzlocha:
Sounds good, I can flip around the equality on `config['anchor_parameters']`. Another alternative which could work is to do the following:

```
assert set(config['anchor_parameters']) <= set(AnchorParameters.default.__dict__.keys())
```

It avoids having to have a hard-coded set of fields however it might be a bit harder to understand. What are your thoughts?

de-vri-es:
Using the keys from the default `AnchorParameters` to validate the `anchor_parameters` subkey sounds like a neat idea.

I'd rather have a more descriptive error than an assertion failure though. We tend to get a high amount of issues here even in cases where the error messages describes exactly what the user did wrong. I don't want to make it worse.

martinzlocha:
Sorry for the delay, I was busy with other work. I have made modifications as discussed.

hgaiser:
Apologies for the delay, but the PR looks good to me, thanks!

PR #1036, opened 2019-06-17
wandb integration, full training and results

syllogismos:
Updated README with full training results on wandb, and links to trained resnet50 and resnet101 checkpoints.

https://app.wandb.ai/syllogismos/keras-retinanet/reports?view=carey%2FObject%20Detection%20with%20RetinaNet

image


hgaiser:
Cool, thanks!

PR #1039, opened 2019-06-20
Added functionality to manually cycle through dataset in debug.py

Raghav-B:
Hi, this is my first pull request so please bear with me!

I've been working with a custom dataset using keras-retinanet and realised that debug.py doesn't automatically cycle through the images in my dataset since by default there is no argument passed into `cv2.waitKey()`, causing it to default to a value a 0. This means that `cv2.imshow()` remains stuck at a single image.

This was easily remedied by increasing the value from 0 to 1000, to create a one second delay between the images - not sure if this was a bug or if it was intended, but the lack of any comments led me to believe the former.

I also ended up adding a simple way to cycle through your dataset using the 'n' and 'b' keys. This should be pretty easy to change to the arrow keys as well.

However I feel the only issue might be the line where I sort `generator.image_names`, as maybe this might run too slow for very large datasets?

I'm not entirely sure about whether this is a concern with regards to debug.py's runtime, as I only sort the list once. However I hope my "feature" is of some use to you!

### DIFF
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -205,13 +205,16 @@ def run(generator, args, anchor_params):
         while True:
             key = cv2.waitKey(1)
             cv2.imshow('Image', image)
-            if key == ord('n'): # press n for next image
+            # press n for next image
+            if key == ord('n'):
                 i += 1
                 break
-            if key == ord('b'): # press b for previous image
+            # press b for previous image    
+            if key == ord('b'):
                 i -= 1
                 break
-            if key == ord('q'): # press q to quit
+            # press q to quit    
+            if key == ord('q'): 
                 return False
     return True
 
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -201,9 +201,18 @@ def run(generator, args, anchor_params):
                 # result is that annotations without anchors are red, with anchors are green
                 draw_boxes(image, annotations['bboxes'][max_indices[positive_indices], :], (0, 255, 0))
 
-        cv2.imshow('Image', image)
-        if cv2.waitKey() == ord('q'):
-            return False
+        key = cv2.waitKey(1)
+        while True:
+            key = cv2.waitKey(1)
+            cv2.imshow('Image', image)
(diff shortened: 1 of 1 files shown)


keras_retinanet/bin/debug.py
```diff
@@ -219,6 +231,9 @@ def main(args=None):
     # create the generator
     generator = create_generator(args)
 
+    # sorting list of images for cycling through dataset
+    generator.image_names = sorted(generator.image_names)
```
hgaiser: What is the reason for sorting it?

keras_retinanet/bin/debug.py
```diff
@@ -201,9 +201,21 @@ def run(generator, args, anchor_params):
                 # result is that annotations without anchors are red, with anchors are green
                 draw_boxes(image, annotations['bboxes'][max_indices[positive_indices], :], (0, 255, 0))
 
-        cv2.imshow('Image', image)
-        if cv2.waitKey() == ord('q'):
-            return False
+        key = cv2.waitKey(1)
```
hgaiser: You can still do `cv2.waitKey()`, so that it will wait indefinitely, right? That eliminates the need of the while loop below.

keras_retinanet/bin/debug.py
Raghav-B: Initially I experienced that the images would cycle through in a random order and so pressing the back key would take me to a random image. It might've been an issue with an earlier implementation of how I modified the i index. 

Turns out there is no need to sort it after all and I can cycle through the images sequentially now, as expected.

keras_retinanet/bin/debug.py
Raghav-B: Yup, I changed this to a single cv2.waitKey(1) and removed the while loop, code is a lot neater now.

hgaiser: I meant `cv2.waitKey()`, without any arguments. If you don't pass any arguments it will just wait idly until a key is pressed, which in our case is sufficient.

Raghav-B: Alright, just made the change

hgaiser:
Thanks!

PR #1043, opened 2019-06-21
cashing images + brightness change

jsemric:
Randomly changing image brightness in Generator. Caching images in CSVGenerator to avoid unnecessary disk IO operations.


jsemric:
need fix travis stuff

PR #1044, opened 2019-06-21
csv image cache + random brightness change

jsemric:
1.) Cashing images in csv_generator.py to avoid unnecessary reading from the disk.

2.) Randomly changing image brightness in generator.py

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -273,6 +273,7 @@ def create_generators(args, preprocess_image):
             args.annotations,
             args.classes,
             transform_generator=transform_generator,
+            cache=args.cache,
             **common_args
         )
 
@@ -280,6 +281,7 @@ def create_generators(args, preprocess_image):
             validation_generator = CSVGenerator(
                 args.val_annotations,
                 args.classes,
+                cache=args.cache,
                 **common_args
             )
         else:
@@ -385,6 +387,7 @@ def parse_args(args):
     csv_parser.add_argument('annotations', help='Path to CSV file containing annotations for training.')
     csv_parser.add_argument('classes', help='Path to a CSV file containing class label mapping.')
     csv_parser.add_argument('--val-annotations', help='Path to CSV file containing annotations for validation (optional).')
+    csv_parser.add_argument('--cache', help='Enable image caching (optional). Only use when the data can fit into memory.', action='store_true')
 
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--snapshot',          help='Resume training from a snapshot.')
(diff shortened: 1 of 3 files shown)


martinzlocha:
It should be possible to enable/disable the cache using a command line parameter from train.py so would be good if you added that. (I'm not a maintainer but that is missing.)

jsemric:
Good point, I'll add it.

keras_retinanet/bin/train.py
```diff
@@ -442,6 +445,7 @@ def main(args=None):
 
     # create the generators
     train_generator, validation_generator = create_generators(args, backbone.preprocess_image)
+    return
```
martinzlocha:
```suggestion
```

hgaiser:
Hey, thank you for the PR. Could you split it up in two PRs though, as they are two very separate topics.

Regarding the brightness:
Could you refactor it to be similar to how we do random transformations, allowing for more random image adjustments. You could make use of the tensorflow functions for [brightness](https://www.tensorflow.org/api_docs/python/tf/image/random_brightness), [contrast](https://www.tensorflow.org/api_docs/python/tf/image/random_contrast), [hue](https://www.tensorflow.org/api_docs/python/tf/image/random_hue), [saturation](https://www.tensorflow.org/api_docs/python/tf/image/random_saturation), and maybe [jpeg quality](https://www.tensorflow.org/api_docs/python/tf/image/random_jpeg_quality).

Regarding the cache:
I'm not so sure about the cache.. in almost all cases imaginable the dataset won't fit in memory. @martinzlocha do you think there is a use case for image cache? What in my opinion would be more interesting, is a tool to "export" the final `inputs` and `outputs` from a generator to one file (.h5 file or something). Then we can add a h5 generator, which loads these files and returns the containing `inputs` and `outputs`. A big caveat for this is that you can't have random transformations, neither visual transformations nor geometrical transformations.

martinzlocha:
Hmmm well memory is becoming much cheaper now and many GPU instances have 64+ GBs of memory. If we assume an image size of 0.5MBs then you can store more than 120k of images in memory which would be sufficient to train RetinaNet so even though it might not be always used it might be useful for someone. And given that it will be disabled by default and doesn't require a lot of additional maintenance (as there are no additional dependencies and the code is short) I think that there isn't a big downside to adding it.

de-vri-es:
Do you have any real world performance measurements for with and without image cache?

I'm asking because the OS also has a disk cache which means subsequent access to the same file will already be much faster without the application doing anything. So I'd really like to see some evidence that it really helps.

jsemric:
Hi, thanks for the response. I'll do another PR with visual effects only as it seems there are not many use cases with the cache feature.

PR #1057, opened 2019-07-06
Visual effects

jsemric:
Adding visual effects in the generator pipeline namely:

1.) brightness change
2.) contrast change
3.) saturation change
4.) hue change

The changes are also made in train.py and debug.py setting the visual effect parameters to slightly transform images.

### DIFF
--- a/keras_retinanet/utils/image.py
+++ b/keras_retinanet/utils/image.py
@@ -198,23 +198,21 @@ def resize_image(img, min_side=800, max_side=1333):
 
 
 def _uniform(val_range):
-    """
-    Uniformly sample from the given range
+    """ Uniformly sample from the given range.
 
     Args
-        val_range: A pair of lower and upper bound
+        val_range: A pair of lower and upper bound.
     """
     return np.random.uniform(val_range[0], val_range[1])
 
 
 def _check_range(val_range, min_val=None, max_val=None):
-    """
-    Check whether the range is a valid range
+    """ Check whether the range is a valid range.
 
     Args
-        val_range: A pair of lower and upper bound
-        min_val: minimal lower bound
-        max_val: maximum upper bound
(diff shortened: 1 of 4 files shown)


keras_retinanet/utils/image.py
```diff
@@ -195,3 +195,170 @@ def resize_image(img, min_side=800, max_side=1333):
...
+        image: Image to clip
+    """
+    return np.clip(image, 0, 255).astype(np.uint8)
+
+
+class VisualEffect:
+    """ Struct holding parameters for applying a visual effect.
+
+    Args
+        contrast_factor:   contrast factor between 0 and 3
```
hgaiser: Could you modify the documentation to be sentences (start with uppercase and end with period). Also, if possible in a few words, try to explain how the values are used to modify the image.

jsemric: I tried it with tensorflow functions following the examples in the `backend` module. However, I had to modify/replace opencv and numpy code in several files. And it didn't work after that anyway. I would have to make more changes I suppose.

I would prefer keeping the current numpy/opencv implementation as it works well with `debug.py` and it would be easier to extend.

jsemric:
Can't see the details of the failed build.

keras_retinanet/utils/image.py
hgaiser: > However, I had to modify/replace opencv and numpy code in several files.

Hmm what changes did you have to make? I would think it just works out of the box.

jsemric: OpenCV functions  (eg., in apply_transform, resize_image which follows visual effect) works with numpy arrays, but tf.image functions return Tensors. So I'd have to replace any numpy/opencv code to tensorflow.

hgaiser:
Thanks again!

PR #1059, opened 2019-07-07
Update README.md

ZFTurbo:
Added link on pretrained weights for keras-retinanet based on open images dataset (OID).


hgaiser:
Very nice, thank you!

PR #1065, opened 2019-07-11
Introduce --display-name, change generator loop (debug.py)

nikostsagk:
Hello to everybody! This is my first pull request ever so try to show some mercy please :P.
This repo is great, so I will try sharing some modifications I am making while working on my project and brought me some  convenience.

To start with, I propose a --display-name parser where displays on the bottom left corner of the image the name of the image. Looks that it works well when the image is in default margins (800x1300), but needs fontScaling for smaller dimensions (e.g. for image 200x200). 

The "for" loop was a bit buggy after key bindings were introduced, so I changed it to "while True".
Also, introduced restrictions in variable "i" so it can take values only between [0, generator.size()].

P.S: sorry about the keybindings change, it was due to different OS.

I am looking forward for your insights,
Cheers

### DIFF
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -219,7 +219,7 @@ def run(generator, args, anchor_params):
             i -= 1
             if i < 0:
                 i = generator.size() - 1
-        
+
         # press q or Esc to quit
         if (key == ord('q')) or (key == 27):
             return False
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -186,7 +186,6 @@ def run(generator, args, anchor_params):
             # resize the image and annotations
             if args.resize:
                 image, image_scale = generator.resize_image(image)
-                print(image_scale)
                 annotations['bboxes'] *= image_scale
 
             anchors = anchors_for_shape(image.shape, anchor_params=anchor_params)
--- a/keras_retinanet/bin/debug.py
+++ b/keras_retinanet/bin/debug.py
@@ -17,6 +17,7 @@ limitations under the License.
 """
 
 import argparse
+import ntpath
 import os
 import sys
 import cv2
@@ -34,7 +35,7 @@ from ..preprocessing.kitti import KittiGenerator
 from ..preprocessing.open_images import OpenImagesGenerator
 from ..utils.keras_version import check_keras_version
 from ..utils.transform import random_transform_generator
-from ..utils.visualization import draw_annotations, draw_boxes
+from ..utils.visualization import draw_annotations, draw_boxes, draw_caption
 from ..utils.anchors import anchors_for_shape, compute_gt_annotations
 from ..utils.config import read_config_file, parse_anchor_parameters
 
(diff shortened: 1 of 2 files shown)


keras_retinanet/bin/debug.py
```diff
@@ -154,6 +155,7 @@ def csv_list(string):
     parser.add_argument('-l', '--loop', help='Loop forever, even if the dataset is exhausted.', action='store_true')
     parser.add_argument('--no-resize', help='Disable image resizing.', dest='resize', action='store_false')
     parser.add_argument('--anchors', help='Show positive anchors on the image.', action='store_true')
+    parser.add_argument('--display-name', help='Display image name on the bottom left corner.', dest='name', action='store_true')
```
hgaiser: I don't think there's a need for the `dest='name'` part, having the argument as `dest_name` will be fine.

keras_retinanet/bin/debug.py
```diff
@@ -200,19 +203,30 @@ def run(generator, args, anchor_params):
...
+
+        ## note that the right and left keybindings are probably different for windows
+        ## press right for next image and left for previous (linux)
+        ## macOS: previous = 'n' | next = 'm'
+
+        if key == 109:
             i += 1
-        # press left for previous image
-        if key == 81:
+            if i == (generator.size() - 1): # Terminate after last image.
```
hgaiser: Hmm good catch. Could you change it to the following:

- Remove the `--loop` argument.
- Change the right action to `i = (i + 1) % generator.size()`
- Change the left action to `if i < 0: i = generator.size() - 1`

That way we have looping with the left / right key.

keras_retinanet/bin/debug.py
```diff
@@ -171,7 +173,8 @@ def run(generator, args, anchor_params):
         args: parseargs args object.
     """
     # display images, one at a time
-    for i in range(generator.size()):
+    i=0
```
hgaiser: An equality sign needs spaces around it :)

keras_retinanet/bin/debug.py
```diff
@@ -200,19 +203,30 @@ def run(generator, args, anchor_params):
...
+            if args.name:	
+                draw_caption(image, [0, image.shape[0]], ntpath.basename(generator.image_path(i)))
+                
         cv2.imshow('Image', image)
         key = cv2.waitKey()
-        # note that the right and left keybindings are probably different for windows
-        # press right for next image
-        if key == 83:
+
+        ## note that the right and left keybindings are probably different for windows
```
hgaiser: Double #

keras_retinanet/bin/debug.py
```diff
@@ -200,19 +203,30 @@ def run(generator, args, anchor_params):
...
-        # press left for previous image
-        if key == 81:
+            if i == (generator.size() - 1): # Terminate after last image.
+                return False
+        if key == 110:
             i -= 1
-        # press q to quit
+            if i < 0:                # Don't let i be less than "0".
+                i = 0
+        ## press q to quit
```
hgaiser: Double #

keras_retinanet/bin/debug.py
```diff
@@ -200,19 +203,30 @@ def run(generator, args, anchor_params):
...
         if key == ord('q'):
```
hgaiser: Not really related to this PR, but could you also add the Escape key here?

keras_retinanet/bin/debug.py
```diff
@@ -17,6 +17,7 @@
 """
 
 import argparse
+import ntpath
```
hgaiser: Prefer to use `os.path`, should be cross platform too.

keras_retinanet/bin/debug.py
nikostsagk: hmm not sure if i understood, like this: 
```'--display-name', help='Display image name on the bottom left corner.', action='store_true'```

hgaiser: Yeah, so without the `dest`

keras_retinanet/bin/debug.py
nikostsagk: nice one!

keras_retinanet/bin/debug.py
nikostsagk: that what i was afraid but indeed, it is cross platform

nikostsagk:
Is it better now?

hgaiser:
Yeah, better :)

there's some linting errors to fix though:
https://travis-ci.org/fizyr/keras-retinanet/jobs/557249482

hgaiser:
Thanks again!

PR #1075, opened 2019-07-17
WIP: Correct computing offset of anchors.

### LINKED ISSUE #1073: under
I'm having trouble understanding the computation done in the beginning of `utils.anchors.shift()`:

```python
```def shift(shape, stride, anchors):
    # shape  : Shape to shift the anchors over.
    # stride : Stride to shift the anchors with over the shape.
    shift_x = (np.arange(0, shape[1]) + 0.5) * stride
    shift_y = (np.arange(0, shape[0]) + 0.5) * stride
```
The goal of this method is to tile the prototype-anchors over a grid of the given shape, and offset the `x1,y1,x2,y2` coordinates of each anchor so they refer to the coordinate system of the original image.

According to this computation, the center of the top-left anchors is `(stride/2, stride/2)` regardless of the given shape or the shape of the original image (and from there the anchors are shifted `stride` pixels apart in every direction).  This seems wrong to me.  

As an example, assume our backbone is a just single 7x7 Conv2D layer, applied to an image of size 200x200 with stride 1, with no padding (e.g. 'valid').  In this case the output of the backbone is 194x194 (with stride 1), and the center of the top-left anchors should be `(3,3)` (because this is the center of the 7x7 window when it is at the top-left of the image).  This is not close to `(stride/2, stride/2)=(0.5, 0.5)`. 

If the padding adds 3 zeros on each direction, and uses a stride of 2 (this is the common first conv layer in ResNet50), then the output is going to be 200x200, and the center of the top-left anchors should be `(0,0)` (and not `(1,1)`).   Even if the stride was larger than 2 the location of the top-left anchors would still be `(0,0)`. 

I would argue that in order to know the correct offset of the top-left anchors, you need to know the shape of the **original image**, not just the layer. Than, if `H, W = image.shape` and `h, w = layer.shape`, the offset should be:  ```(H - (h - 1)*stride) / 2.0, (W - (w-1)*stride) / 2.0```

What do you think?

hgaiser:
This PR fixes the computation of the offset of the anchors, thanks to @jonilaserson . This PR fixes #1073 .

**This affects all networks trained before this change. They should be retrained after this is merged for it to work with the latest version.**

ps. I think we should hold off merging this, until we have a retrained version of the COCO weights.


keras_retinanet/utils/anchors.py
```diff
@@ -234,24 +234,30 @@ def anchors_for_shape(
...
+        image_shape    : Shape of the input image.
+        features_shape : Shape of the feature map.
+        stride         : Stride to shift the anchors with over the image.
+        anchors        : The anchors to apply at each location.
     """
 
+    # compute the offset of the anchors based on the image shape and the feature map shape
+    # see https://github.com/fizyr/keras-retinanet/issues/1073 for more information
+    offset_x = (image_shape[1] - (features_shape[1] - 1) * stride) / 2
+    offset_y = (image_shape[0] - (features_shape[0] - 1) * stride) / 2
```
jonilaserson: If you want to support python 2.X, better divide by 2.0

hgaiser: Good catch! I wish we weren't supporting 2.x, but unfortunately we are :p

vcarpani (approved):
LGTM, thanks @jonilaserson for catching this :1st_place_medal: 

@hgaiser before merging let's wait for COCO to be trained both for keras-retinananet and keras-maskrcnn. We should also properly announce this change on the slack channel and bump the release to 0.6.0.

hgaiser:
I've also modified the tensorflow `shift` function. Unfortunately we need this function twice, once in numpy and once in tensorflow ops. The evaluation on COCO, using the same weights as before, give the following results:

```
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.286
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.530
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.280
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.170
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.336
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.368
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.257
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.424
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.464
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.315
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.523
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.567
```

The original results:

```
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.350
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.537
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.374
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.191
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.383
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.472
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.306
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.491
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.533
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.345
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.577
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.681
```

There's a difference there, which makes sense. The difference for IoU=0.5 is rather small, but the difference for IoU=0.75 is large (causing the difference in mAP to be relatively large as well). Basically all detections have been shifted a few pixels to the top left corner, which isn't that critical with a lenient overlap threshold of 0.5, but for 0.75 it is more severe.

Next I'll start a training process to finetune the network for this new setting, and add unit tests for this function.

hgaiser:
Added unit tests and started training starting from the previous best weights. The first epoch moved the mAP from 0.286 to 0.333, so that's a nice first increase. I'll see how far it gets, hopefully it becomes at least as accurate as the original model :)

I'm awaiting the results of training (and that of keras-maskrcnn) before merging this. We'll have to tag it as a new version as well.

hgaiser:
This is still on my radar btw, but it takes a while to retrain. So far the best mAP I got was 0.343, I'm trying to see if I can push it closer to 0.350 (with simply more training).

jonilaserson:
Thanks for the update, Hans!

It should be the same, right?  For models that use ResNet50 as backbone the new computation gives the same result as the previous computation.  The difference would occur for models that play with the padding or strides differently.

hgaiser:
> Thanks for the update, Hans!
> 
> It should be the same, right? For models that use ResNet50 as backbone the new computation gives the same result as the previous computation. The difference would occur for models that play with the padding or strides differently.

Yeah it should be, but the 0.350 mAP score from before was basically luck. I never succeeded in reproducing that exact score ;) a score of 0.34 is fine, but there was only one snapshot that reached 0.350.

martinzlocha:
Is there a reason why you are not using seeds for training?

hgaiser:
Very good question: no there's not :D

Too late now ;)

martinzlocha:
😢 Would it be worth creating a PR which allows users to set seeds using a parameter and by default they aren't set?

hgaiser:
Yeah, I would think it can be useful, definitely.

hgaiser:
I've started training from scratch, but after a while the training stopped improving (achieved around 0.27 mAP). I've restarted training multiple times (starting from the weights of the previous iteration) and have pushed the mAP to 0.292 so far.

This annoys me a bit because it suggests that the training procedure is not working properly (which I've known for a while now). Perhaps a different learning rate scheduler, or different optimizer, or both? I don't have time to look into these settings, so if someone wants to help then it is much appreciated.

hgaiser:
For the record, this PR is still valid, but I haven't got the time to properly train a new model that gets similar accuracy as the current model.

ZFTurbo:
Have a question about this PR. Could this PR potentially improve models accuracy after we additionally train them with correct data?

hgaiser:
I would suspect, considering it's only off by a few pixels, that the network learns to compensate for this offset. Meaning it shouldn't matter much. On the other hand, the amount of offset is affected by the image size so it is never a constant offset, which could confuse the network. Bottom line, I don't know :)

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #1079, opened 2019-07-18
Fix 'draw_detections'

nikostsagk:
`predict_on_batch` returns 300 boxes by default (unless specified otherwise), and then these are chopped down by `score_threshold=0.05` and `max_detections=100` by default.

However, when it comes to drawing by `draw_detections` the default `score_threshold` is 0.5 leading in wrong visual estimation.

This PR inserts in the function the correct `score_threshold` specified by the training process.


hgaiser:
Thank you!

PR #1085, opened 2019-07-25
Add the option to seed training.

martinzlocha:
As we briefly discussed in #1075 I am adding the possibility to provide seed values.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -36,3 +36,4 @@ please create an issue or pull request at https://github.com/fizyr/keras-retinan
 * Eduardo Ramos
 * DiegoAgher
 * Alexander Pacha
+* Martin Zlocha
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -21,9 +21,11 @@ import os
 import sys
 import warnings
 
+import random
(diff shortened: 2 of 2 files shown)


hgaiser:
I don't think this will work together with how we generate a random transformation:
https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/utils/transform.py#L286

martinzlocha:
Sorry for the delay, hope its fine now.

hgaiser:
It seems fine. Did you test it somehow to make sure the results were reproduced over two runs?

de-vri-es:
I'm not sure if this can work correctly with multithreaded workers running the generator. If variation in processing time changes the way work is distributed over threads, that would affect the random transforms.

With a single worker it should be fine I think. Might be worth adding a warning if multiple workers are used together with a seed.

hgaiser:
> With a single worker it should be fine I think. Might be worth adding a warning if multiple workers are used together with a seed.

@martinzlocha what do you think?

martinzlocha:
@hgaiser Yes I completely agree with the multithreading issue. Unfortunately I don't have the time right now to write proper tests and not sure when I will come back to this so it might be best to close the PR if its not a requested feature.

PR #1096, opened 2019-08-09
Add a section about anchor optimization

martinzlocha:
As we have discussed on Slack I have added a section about anchor optimization, which links to my repository with the actual code and further instructions. Let me know if I can improve it somehow :)


hgaiser:
Thank you, very nice addition!

PR #1097, opened 2019-08-12
Update image.py

Yuliang-Huang:
add image loading method for dicom file


KimWanki:
Thanks with your dicom file loader.
I have a question.
I want to train my model with another input size.
ex) (batch_size, h, w , channel = 4,5,6,7...) 
is there any way to change the size of channels?

PR #1099, opened 2019-08-16
Updates to debug.py

jnmaloof:
Two small updates to debug.py that make my life easier and might be nice for others:

1) Automatically adjust forward/back keys to "m" and "n" if on Mac

2) Create a "--no-gui" mode, whereby instead of opening an image browser debug will loop through each image and write out a .png file  with the annotations to the current directory.


de-vri-es:
Thanks for the PR! In general I think the changes are nice. Some comments though.

How about just adding the n and m keys as back/forward, disregarding the OS? So just use left/right keys and n and m on all platforms. Also, what's the issue with the current keybindings on OS X? If they're broken, can we fix those somehow?

Saving image to the current working directory could be problematic when there are multiple images with the same name in different subfolders. Most datasets will contain relative paths, so for those we could just take the whole path instead of the basename. For absolute paths we could then either give an error, or skip them, or take the basename anyway.

Also, debug.py.bak should be removed :)

jnmaloof:
Thanks for the feedback.  I'll work on your suggestions.  I don't know what is up with OSX key bindings, but the left and right arrow keys do not return any keycode when using `cv2.waitKey()`.

hgaiser:
@jnmaloof I believe `cv2.waitKey` only returns ASCII characters, which is why it doesn't return left/right keys. I'm not sure if it returns non-ASCII characters in some cases though. Could you look into using `cv2.waitKeyEx`?

de-vri-es:
Still not sure about saving all debug images in the same folder. I'd really like to keep the folder structure of the original dataset.

That just leaves the question how to deal with absolute paths. Could even just do `os.path.join('.', absolute_path)` for those weird cases where datasets have absolute paths.

jnmaloof:
@de-vri-es Oh, yes, sorry about that.  A but under the weather this week and spaced out on that.  How about I do as you suggest (place in same folder(s) as original images) but also provide an optional argument to override, where the use can specify an alternative output directory?

de-vri-es:
An `output-dir` option would probably be best. It could default to "." (current working directory), and then append the path from the dataset with "_debug.png".

I was thinking something along these lines:
```
_, path = os.path.splitdrive(image)
if os.path.isabs(path)
  path = path.relpath(path, '/')
path = os.path.join(output_dir, path)
# add `_debug.png` like before
```

jnmaloof:
I think the default should be to put the annotated image in whatever
directory the original image was in.  I don't see what absolute vs relative
paths is an issue; whether the user provided an absolute or relative path
to the original image, so long as we use the path that they provided (and
add the "debug.png") it should go to the right place, correct?

de-vri-es:
I'm not so sure. If we always put the debug images next to the originals it would be somewhat difficult to create two separate sets of processed images from the same dataset. And depending on the situation, a user may not even have write permission on the dataset folders.

@hgaiser: what do you think?

jnmaloof:
Sure, I'd be happy for default to be current working directory.

de-vri-es:
Right, but then we do need to deal with absolute paths specially. Otherwise `--output-dir` wouldn't work for datasets with absolute paths. `os.path.join('/output/dir', '/absolute/path')` just gives `/absolute/path`.

jnmaloof:
Trying to understand your comment.  I was thinking that output-dir would be
non-hierarchical and would just contain all of the "_debug.jpg" files.
Under that scenario absolute vs relative doesn't matter because I would
just take the actual file name of the input files, modify it and place that
in output-dir.  I take that you are thinking of something else, i.e.
preserving path hierarchy leading to the input images, but placing that
hierarchy within output-dir?

I think that I could check for whether or not the path is absolute, and if
it is then use os.path.relpath(imagepath, startpath="/") or
PurePath.relative_to("/") to convert the absolute path to a relative one
and append that to the output-dir.

I will note that personally, that is not what I as a user would want (I
prefer the simpler behavior that I initially envisioned), but my data sets
may be simpler than most.  So if preserving the input image file hierarchy
in output-dir is what you think is best I can work on it.

de-vri-es:
I'm mainly worried about cases where a dataset contains images with the same name in different subfolders. If all debug images go to the same folder, they'll be overwritten.

jnmaloof:
got it.  I'll work up something.  I'll probably add a flag option to
flatten the hierarchy in case that is the user's preference.  I'm on
vacation for a couple of weeks so it may be a little while.

hgaiser:
>  I will note that personally, that is not what I as a user would want (I prefer the simpler behavior that I initially envisioned), but my data sets may be simpler than most. So if preserving the input image file hierarchy in output-dir is what you think is best I can work on it.

We used to have that in other places as well, but as @de-vri-es said there are datasets where the filenames are identical, but they are placed in different dirs. Outputting them in one directory means files will be overwritten. The best way to work around this is to preserve file hierarchy unfortunately.

de-vri-es:
Note that if your dataset is a flat directory, your output will be too. So for custom datasets you can already get the behaviour you want. Still, a flag to flatten the hierarchy seems nice to have.

jnmaloof:
I am on board with this.  Flatten flag will still be nice because it will
remove directory tree above the files.

I'll update in late September.  Trying to stay off of the computer as much
as possible while on holiday.

jnmaloof:
What do you think about using `os.path.commonpath(paths)` ?  The description is:

> Return the longest common sub-path of each pathname in the sequence paths. Raise ValueError if paths contains both absolute and relative pathnames, or if paths is empty. Unlike commonprefix(), this returns a valid path.

This will trim the extraneous parts of the file path.  (And will completely flatten it if all images are in a single directory, so I won't need to give a "--flatten" flag).

jnmaloof:
To clarify my comment about `os.path.commonpath(paths)`, I would use that command to find the common (and therefore extraneous with respect to preventing clobbering) part of the path and then remove that from each files path.

jnmaloof:
OK, I have implemented what I suggested above: any common part of the directory path is stripped, so if all of the images are in one directory, then they are output into a single directory.  If they are in multiple directories, that is preserved, starting with the unique parts of the path.

I have only tested this on the `csv`  data set type, but so long as the other types have a `generator.image_path()` method (which they must since that method was already used), then they should work.

I have also only tested this on a Mac.  I use the `os.*` functions for path manipulation so it should work on other platforms.

jnmaloof:
OK it would be helpful at this point if you shared how you do want it to
work, because clearly my ideas are not getting there.  Based on your
feedback, one possibility is to have the default behavior be for the
images_debug.png files to go back into the same directory that the input
images came from. Does that work for your workflow?  And then I could add
flags for changing that to make it more like I would like.  Or do you want
the default for the full hierarchy to be placed within the output directory
(which by default would be the directory where the program was called
from)?  It just seems a little awkward for the default to create something
like:

`/Users/jmaloof/imageAnalysis/debugimages/Users/jmaloof/imageAnalysis/inputimage/image001_debug.png`

which is what would happen if I don't trim off the common path but do use
the working directory as root of the output.

As an aside, I note that this is new functionality; currently these debug
image files are not created at all.  So it shouldn't really be messing up
anyone's pipeline.  (Although I can see that it would make it hard to
incorporate these new files into an existing pipeline).

de-vri-es:
> OK it would be helpful at this point if you shared how you do want it to work, because clearly my ideas are not getting there.

Well, I suggested something earlier:
```py
_, path = os.path.splitdrive(image)
if os.path.isabs(path)
  path = path.relpath(path, '/')
path = os.path.join(output_dir, path)
# add `_debug.png` like before
```

I still like this approach.


> `/Users/jmaloof/imageAnalysis/debugimages/Users/jmaloof/imageAnalysis/inputimage/image001_debug.png` which is what would happen if I don't trim off the common path but do use the working directory as root of the output.

I expect most datasets to use relative paths and be self-contained. If they are, appending to the current working directory gets you something like `./001/image_debug.png`, which isn't nearly as horrible as what you describe.

I agree that datasets with absolute paths will end up rather nasty, but I think that's still better than always putting them in the same folder as the original image or always flattening the hierarchy.

If we want a real nice experience for datasets with absolute paths, we could add an option like `--strip-output-prefix /some/path`. But personally, I think it's enough to do something technically correct for absolute paths without making it the best of experiences. After all, datasets with absolute paths are a bit weird.

Do keep in mind that absolute paths on windows can also contain drive letters, which require special handling. On windows, each drive has it's own working directory and you can have relative paths for different drives. (Yeah, windows is weird.)


> And then I could add flags for changing that to make it more like I would like.

Yes, by all means. I like the idea of `--flatten-output` or similar.

de-vri-es:
I'd like to get this merged, so I tried to address the remaining feedback in #1145. If that is merged, I think this will automatically be listed as merged by github too, since all commits in this PR are also contained in that PR.

PR #1125, opened 2019-10-03
Update required Keras version to 2.3.0

AndreaPi:
In accordance to https://github.com/fizyr/keras-retinanet/commit/d1541918186b1ee99e75667b51176c19639bd979


de-vri-es:
Good catch, thanks!

PR #1127, opened 2019-10-05
Updated Train.py to fix tensorboard issue

KalraA:
Moving the tensorboard callback to the end of the callback list allows mAP and LR to be updated when the epoch ends. As of right now they are 1 epoch behind in the tensorboard graph.


hgaiser:
I see, thank you !

KalraA:
Can you add me to the contributors list? 
Agastya Kalra - agastya@akasha.im

hgaiser:
Could you make a PR for that?

PR #1130, opened 2019-10-10
Fixes for tensorflow 2 compatibility

hgaiser:
This PR should fix compatibility with tensorflow 2.


hgaiser:
Hmm travis still doesn't seem to want to build it. Anyone some thoughts?

vcarpani:
LGTM, merging

de-vri-es:
@hgaiser: Do the test pass locally? And if so, after how much time?

Also, does this mean it's all slower with tensorflow 2.0 than it was with 1.x? That'd be kind of a shame.

PR #1136, opened 2019-10-11
move github templates

Borda:
it is a standard location after Github updated this feature...


Borda:
pls consider also add a bage of code quality like   [CodeFactor](https://www.codefactor.io/repository/github/borda/keras-retinanet)

de-vri-es:
Moving ISSUE_TEMPLATE seems like a good idea.

However, CONTRIBUTORS.md doesn't really have anything to do with github. Our license requires that is is redistributed along with keras-retinanet itself. It should definitely be in the root of the repository.

Regarding a code quality badge, I'm personally not a huge fan of automated code quality assessment. On the other hand, apparently it gives the repo an A, so I'm not opposed to it either.

Borda:
ok, let's keep `CONTRIBUTORS.md` in the root
there is another point regarding issue template, there is an option to separate feature requests and bug report (having different templates, see e.g. [here](https://github.com/Borda/BIRL/tree/master/.github/ISSUE_TEMPLATE)) are interested in it?

de-vri-es:
Hmm, yes, that does seem interesting. Thanks for pointing this out :)

Our current template isn't really a template but a message people should read before filing bug reports. We could use what we have as bug report template, and just use an empty feature request template.

Do you want to do implement that in this PR?

Borda:
OK ai will prepare it and then let you know...

hgaiser:
Thank you, looks good to me. I like the separation of bug and PR template, makes sense.

PR #1141, opened 2019-10-14
add --no-resize parameter to allow training w/o resizing input images

AndreaPi:
Add a parameter `--no-resize` to `train.py`, so that one can train a model without any resizing of the input images.  The mechanism is very simple: the method `resize_image(self, image)` of class `Generator(keras.utils.Sequence)` is modified so that if the `no_resize` parameter is passed when initializing a `generator` object, then `resize_image` always returns a scale of 1, instead of computing it.

```
    def resize_image(self, image):
        """ Resize an image using image_min_side and image_max_side.
        """
        if self.no_resize:
            return image, 1
        else:
            return resize_image(image, min_side=self.image_min_side, max_side=self.image_max_side)
```

Since the change is made in  the `Generator` class, it works for all data generators, since they inherit from this class.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -215,6 +215,7 @@ def create_generators(args, preprocess_image):
         'config'           : args.config,
         'image_min_side'   : args.image_min_side,
         'image_max_side'   : args.image_max_side,
+        'no_resize'        : args.no_resize,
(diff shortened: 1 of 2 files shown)


hgaiser:
Alright, nice, I see a use case for this. Thanks!

Issue #1113, opened 2019-09-13
debug.py script not working correctly with COCO-type datasets


AndreaPi:
I have a custom dataset, which uses COCO annotations, and I want to check it using the `debug.py` script.  On a Mac, I'm launching the script as

```
/somepath/.local/bin/retinanet-debug --display-name --annotations --anchors coco /someotherpath/COCO
```

The script is nearly identical to the master version: I only changed

```
        if key == 83:
            i = (i + 1) % generator.size()
        if key == 81:
```

to

```
        if key == ord('m'):
            i = (i + 1) % generator.size()
        if key == ord('n'):
```

because I couldn't, for the love of it, find what the heck keys 83 & 81 correspond to on my keyboard.

Anyway, as soon as I press key `m` to go to the next image, I get this error:

```
Traceback (most recent call last):
  File "/somepath/.local/bin/retinanet-debug", line 10, in <module>
    sys.exit(main())
  File "/somepath/.local/lib/python3.7/site-packages/keras_retinanet/bin/debug.py", line 269, in main
    run(generator, args, anchor_params=anchor_params)
  File "/somepath/.local/lib/python3.7/site-packages/keras_retinanet/bin/debug.py", line 221, in run
    draw_caption(image, [0, image.shape[0]], os.path.basename(generator.image_path(i)))
AttributeError: 'CocoGenerator' object has no attribute 'image_path'
```

Looks like the `debug.py` is expecting the `CocoGenerator`object to have some attribute that it doesn't have. Can you help me?

hgaiser:
Seems that the argument `--display-name` is using an attribute that is not part of the base generator class. Could you make a PR to resolve this? The best way would be to add a base function in generator.py, `image_path`, and add this function to the different generators.

AndreaPi:
@hgaiser thanks for the tip! I'm at a conference this week, but once I'm back I'll have a look at the code and let you know if I can do it. The main issue I can see with my first PR is that I'm not familiar with testing/CI conventions for the `keras-retinanet` source, thus I don't know what I would need to include (apart from bare,  functional code), in order for it to be an acceptable PR.

de-vri-es:
It's okay to just submit a PR and let the CI server do it's thing. It will report problems automatically. Those problems can be fixed as they're reported. It's slower than running the tests locally, but there's no rush :)

And in the mean time you should be able to use your patched version locally to use the debug script with COCO.

AndreaPi:
[...]

PR #1143, opened 2019-10-15
fix issue #1113, but only for the COCO generator

AndreaPi:
This fixes the issue https://github.com/fizyr/keras-retinanet/issues/1113. However, I wasn't able to make it work at the `Generator` class level, thus I implemented the fix in the `CocoGenerator` class. I just add the lists `images` and `image_names` in the `__init__` method, and an `image_path` method which is based on the same method in the `CSVGenerator` class. I tested it on my custom dataset in COCO format, and it works.

### DIFF
--- a/keras_retinanet/preprocessing/coco.py
+++ b/keras_retinanet/preprocessing/coco.py
@@ -40,8 +40,6 @@ class CocoGenerator(Generator):
         self.set_name  = set_name
         self.coco      = COCO(os.path.join(data_dir, 'annotations', 'instances_' + set_name + '.json'))
         self.image_ids = self.coco.getImgIds()
-        self.images    = self.coco.loadImgs(self.image_ids)
-        self.image_names = [image['file_name'] for image in self.images] 
 
         self.load_classes()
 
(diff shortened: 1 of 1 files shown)


de-vri-es:
Thanks for the PR. I'll try to get round to reviewing in the weekend.

hgaiser:
The addition itself is good, but there I would really prefer to have it in all generators before merging it. Is that something you can work on @AndreaPi ?

Also, while we have it, we might as well use it in `CocoGenerator.load_image`.

AndreaPi:
@hgaiser I don't think so. I'm not familiar with formats other than COCO (which works with this fix) and csv (which works already, its generator has the `image_path`). I would have to learn about the formats (KITTI, OID, PASCAL VOC), download the datasets (I guess they're all Open Access, right?), modify the generators and test them. I don't think I'll be able to do it this year.

No problem for your other request!

> Also, while we have it, we might as well use it in `CocoGenerator.load_image`.

I'll modify the code so that `CocoGenerator.load_image` uses `CocoGenerator.image_path`, and push another commit to this same fork.

AndreaPi:
Done! I modified CocoGenerator.load_image, so that now it uses CocoGenerator.image_path.

hgaiser:
I'll see if I can add the same functionality to the other generators in a different PR. Thanks!

PR #1146, opened 2019-10-21
Disable tf2 behavior.

hgaiser:
Since fully migrating to tf2 is going to be a bit difficult, the intermediate step is to disable v2 behavior. Unfortunately there is a bug in keras which prevents us from properly disabling v2 behavior, but it should hopefully be fixed with [this](https://github.com/keras-team/keras/pull/13476) PR.

Hopefully we can remove this at some point in the future, but for now let's get things working as they were using tensorflow 2.

ps. unit tests are still expected to fail due to the bug in Keras, but they should at least finish now. Let's wait merging this PR until the unit tests seem okay (aside from the Keras bug).

### DIFF
--- a/keras_retinanet/backend/tensorflow_backend.py
+++ b/keras_retinanet/backend/tensorflow_backend.py
@@ -17,6 +17,12 @@ limitations under the License.
 import tensorflow
 
 
+def disable_tensorflow_v2_behavior():
+    """ See https://www.tensorflow.org/api_docs/python/tf/compat/v1/disable_tensorflow_v2_behavior .
+    """
+    tensorflow.compat.v1.disable_v2_behavior()
+
+
 def ones(*args, **kwargs):
     """ See https://www.tensorflow.org/api_docs/python/tf/ones .
     """
--- a/keras_retinanet/bin/convert_model.py
+++ b/keras_retinanet/bin/convert_model.py
@@ -27,9 +27,14 @@ if __name__ == "__main__" and __package__ is None:
     __package__ = "keras_retinanet.bin"
 
 # Change these to absolute imports if you copy this script outside the keras_retinanet package.
+from .. import backend
 from .. import models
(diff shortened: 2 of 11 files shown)


hgaiser:
Alright, the tests failed as expected, feel free to merge if you approve this PR @vcarpani .

keras_retinanet/backend/tensorflow_backend.py
```diff
@@ -17,6 +17,12 @@
 import tensorflow
 
 
+def disable_v2_behavior():
```
de-vri-es: I know we only support tensorflow, and it's extremely unlikely that another back-end will ever be supported, but: `keras.backend` is meant to abstract away implementation details of backends rather than expose them. That's also why you still need to import tensorflow manually if you want to use things from it directly.

In that light, it may be better not to call this `disable_v2_behavior`, but `disable_tensorflow_v2_behavior`. That still exposes backend details, but not under a generic sounding name.

Not extremely important, but I thought I'd mention it.

hgaiser: Yep, not really important, but you're right :+1:

keras_retinanet/bin/convert_model.py
```diff
@@ -27,11 +27,16 @@
...
 
 # Change these to absolute imports if you copy this script outside the keras_retinanet package.
+from .. import backend
 from .. import models
 from ..utils.config import read_config_file, parse_anchor_parameters
 from ..utils.gpu import setup_gpu
 
 
+# Disable Tensorflow 2 behavior as we experience issues with it.
+backend.disable_tensorflow_v2_behavior()
```
vcarpani: Maybe:
```
if keras.backend.backend() == 'tensorflow': backend.disable_tensorflow_v2_behavior()
```

hgaiser:
Seems like the Keras PR will never be merged, so let's just merge this PR at least.

PR #1153, opened 2019-10-22
opencv-python -> opencv-python-headless

MichaelMonashev:
opencv-python-headless package do not contain any GUI functionality. It is smaller and suitable for more restricted environments.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ h5py
 keras
 matplotlib
 numpy>=1.14
-opencv-python>=3.3.0
+opencv-python-headless>=3.3.0
 pillow
 progressbar2
 tensorflow


de-vri-es:
Thanks for the PR!

It would be nice to easily allow using keras-retinanet with a headless opencv. However, there are parts of the project that use the GUI functionality of opencv. Most notably probably `debug.py` that shows a graphical window.

I'm not sure if python has a nice way to specify optional dependencies. If so think we would probably still want the default to be opencv with GUI functionality, but have an option to replace it with a headless opencv. I'm not sure how to do that though.

MichaelMonashev:
I think the optional dependencies makes installation more difficult for users.

Issue #1149, opened 2019-10-22
AttributeError: module 'tensorflow.compat' has no attribute 'v1'


513221019:
when I trained on only one image， something was wrong。
 
 return tensorflow.compat.v1.image.resize_images(images, size, methods[method], align_corners)

AttributeError: module 'tensorflow.compat' has no attribute 'v1'

who can help me ？ thanks

de-vri-es:
Looks like adding compatibility with tensorflow 2 may have broken compatibility with tensorflow 1. Not sure if the intention is to stay compatible with tf1 for now. If so, we should probably check for the e old `tf1` function before trying `compat.v1` (or vice-versa).

@hgaiser, @vcarpani: is it the intention to still be compatible with tensorflow 1 for now?

hgaiser:
I'm in favor of supporting tensorflow 1, but only 1.14 and 1.15 as they already have the same API as 2. Maybe it would be better to add a version check for tensorflow in `tensorflow_backend.py`, to ensure 1.14 or higher.

PR #1157, opened 2019-10-30
Bump minimum tensorflow version to 1.14, and check in all scripts.

de-vri-es:
This PR bumps minimum tensorflow version to 1.14 and calls the check in all scripts, as suggested at https://github.com/fizyr/keras-retinanet/issues/1149#issuecomment-545888936


de-vri-es:
Note that I kinda agree with your comment that it should be in `backend/tensorflow_backend.py`, but you added `utils/tf_version.py` :)

hgaiser:
Yep indeed, we should focus on supporting tensorflow 1.14+ as it is forward compatible with tensorflow 2. Thanks :+1:

PR #1158, opened 2019-10-30
Added support for SE and EfficientNet backbones

ZFTurbo:
Added support for following pre-trained backbones:
'seresnet18', 'seresnet34', 'seresnet50', 'seresnet101', 'seresnet152', 'seresnext50', 'seresnext101', 'senet154', 'EfficientNetB0', 'EfficientNetB1', 'EfficientNetB2', 'EfficientNetB3', 'EfficientNetB4', 'EfficientNetB5', 'EfficientNetB6', 'EfficientNetB7'

It has dependency on following python modules: '[classification_models](https://github.com/qubvel/classification_models)' and '[efficientnet](https://github.com/qubvel/efficientnet)'. But they will be called only if user will use these backbones.

### DIFF
--- a/keras_retinanet/models/effnet.py
+++ b/keras_retinanet/models/effnet.py
@@ -48,7 +48,6 @@ class EfficientNetBackbone(Backbone):
         file_name = model_name + '_weights_tf_dim_ordering_tf_kernels_autoaugment_notop.h5'
         file_hash = WEIGHTS_HASHES[model_name][1]
         weights_path = get_file(file_name, BASE_WEIGHTS_PATH + file_name, cache_subdir='models', file_hash=file_hash)
-        # print('Image net weights loaded: {}'.format(weights_path))
         return weights_path
 
     def validate(self):
--- a/keras_retinanet/models/effnet.py
+++ b/keras_retinanet/models/effnet.py
@@ -1,0 +1,149 @@
+"""
+Copyright 2017-2018 Fizyr (https://fizyr.com)
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


ZFTurbo:
I currently experimenting with these backbones. I was able to get 0.40 mAP with SE-Resnext50 with freezed backbone. Which is worse than best achieved 0.49 mAP for resnet152. Currently I try to investigate why mAP became worse when I'm unfreeze backbone. Freeze of BN layers didn't help.

keras_retinanet/models/effnet.py
```diff
@@ -41,7 +41,15 @@ def retinanet(self, *args, **kwargs):
...
         """
-        return
+        from efficientnet.model import BASE_WEIGHTS_PATH
+        from efficientnet.model import WEIGHTS_HASHES
+
+        model_name = 'efficientnet-b' + self.backbone[-1]
+        file_name = model_name + '_weights_tf_dim_ordering_tf_kernels_autoaugment_notop.h5'
+        file_hash = WEIGHTS_HASHES[model_name][1]
+        weights_path = get_file(file_name, BASE_WEIGHTS_PATH + file_name, cache_subdir='models', file_hash=file_hash)
+        # print('Image net weights loaded: {}'.format(weights_path))
```
vcarpani: Forgot to remove debud 'print' ;)

vcarpani:
> I currently experimenting with these backbones. I was able to get 0.40 mAP on Open Images Dataset with SE-Resnext50 with freezed backbone. Which is worse than best achieved 0.49 mAP for resnet152. Currently I try to investigate why mAP became worse when I'm unfreeze backbone. Freeze of BN layers didn't help.

Do you have any comparison between ResNext 50 and Resnet 50?

ZFTurbo:
I was able to get 0.39 mAP for resnet50 before unfreeze. So SE-ResNext with 0.40 is a little bit better. But overall I got 0.44 mAP for ResNet50 (after unfreeze).

vcarpani:
LGTM, I am merging so also other users can start testing with different backbones ;)

PR #1160, opened 2019-10-31
Add image_path methods to KITTI and Pascal VOC datasets.

hgaiser:
@AndreaPi this should add `image_path` to KITTI and Pascal VOC datasets (and also add it to the abstract Generator class).


vcarpani:
LGTM, merging.

PR #1169, opened 2019-11-05
Implemented F1 score for assessments.

SalahAdDin:
I'm working on that but it seems i don't understand how to get that value, so, if you can help me, it would be great, thanks.

### DIFF
--- a/keras_retinanet/bin/evaluate.py
+++ b/keras_retinanet/bin/evaluate.py
@@ -150,7 +150,7 @@ def main(args=None):
         from ..utils.coco_eval import evaluate_coco
         evaluate_coco(generator, model, args.score_threshold)
     else:
-        average_precisions = evaluate(
+        average_precisions, f1_scores = evaluate(
             generator,
             model,
             iou_threshold=args.iou_threshold,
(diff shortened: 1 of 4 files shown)


hgaiser:
Thanks for the PR! I'm not sure about the use case for computing the F1 score though. All object detection datasets I know use some form of mAP to report there accuracy. Why do you think the F1 score should be added?

SalahAdDin:
Because, if you review most of academic articles, they no only use mAP, they also use F1 score to compare the performance along the different deep learning techniques.

I could add here all articles that use it, also my thesis advisor requires from me F1 score; it is, as far as i know, a standar in the computer sciences.

This implementation is not perfect and it is nor working: F1 score should be between 0.0 to 1.0, but i'm getting values over 1.0.

Thanks.

SalahAdDin:
All help is welcome.

SalahAdDin:
@hgaiser Did you implement _F1 score_ in your _py-faster-rcnn_  implementation?

hgaiser:
Could you also update the code [here](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/callbacks/eval.py#L63) ?

> @hgaiser Did you implement _F1 score_ in your _py-faster-rcnn_ implementation?

What py-faster-rcnn implementation do  you mean?

SalahAdDin:
Done.

I got it wrong, it was commented here: https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/utils/eval.py#L33

hgaiser:
Thank you for the PR :)

SalahAdDin:
@MiguelAngelL07 You can add it making a PR. Please.
@hgaiser Did you review this? it was not working for me,

SalahAdDin:
I found the definition of [**F1**:](https://arxiv.org/pdf/1807.11809.pdf)  

> The harmonic mean of precision and recall. For multi-class
> classification problems, F1 gets averaged among all the classes. It is
> mentioned as F-measure in (Minh, et al., 2017).
> F1=2 * (TP*FP) / (TP+FP)

So, I implemented this wrong.

hgaiser:
@SalahAdDin can you make a PR to correct it?

SalahAdDin:
@hgaiser I don't have time to do it now, i'll appreciate your help here.

hgaiser:
> @hgaiser I don't have time to do it now, i'll appreciate your help here.

Unfortunately I am also quite limited in my time.

PR #1172, opened 2019-11-06
Small fixes for new EfficientNet and SEResNet backbones

ZFTurbo:
Removed unnecessary dependence on keras_resnet for EfficientNet and SEResNet backbones. It also fixes problem with loading saved snapshots.


hgaiser:
Thanks for the fixes @ZFTurbo .

PR #1178, opened 2019-11-10
added initial-epoch to argparse`

bfialkoff:
Can be used together with weights to conveniently continue training from a snapshot


keras_retinanet/bin/train.py
```diff
@@ -412,6 +412,7 @@ def csv_list(string):
     parser.add_argument('--gpu',              help='Id of the GPU to use (as reported by nvidia-smi).')
     parser.add_argument('--multi-gpu',        help='Number of GPUs to use for parallel processing.', type=int, default=0)
     parser.add_argument('--multi-gpu-force',  help='Extra flag needed to enable (experimental) multi-gpu support.', action='store_true')
+    parser.add_argument('--initial-epoch',    help='Epoch from which to begin the train, useful if resuming with weights.', type=int, default=0)
```
hgaiser: Minor detail, but shouldn't this be `if resuming from snapshot.`?

bfialkoff:
Ostensibly yes, that would be the most prevent scenario. Shall I update?

hgaiser:
Yes please :)

de-vri-es:
It looks like you accidentally added IDE files in the last commit. Could you remove those?

I kinda preferred "useful" over "use" btw.

bfialkoff:
oops, yeah i'm on it.

hgaiser:
I see you also added a `makedirs` for the tensorboard dir, seems to make sense.

Thanks for the contribution!

AndreaPi:
This was a great addition (thanks!), but the new API might be a bit confusing, IMO.  Right now, imagine you ran your model for 20 epochs, and you want to run it for another 20 epochs. if you do something like

```
retinanet-train --batch-size=1 --steps=50 --epochs=20 --initial-epoch=21 --snapshot /snapshots/resnet50_coco_best_v2.1.0.h5 --compute-val-loss --tensorboard-dir /snapshots/logs/$(date +%m-%d-%H%M) --snapshot-path /snapshots coco /data/COCO/
```

the code will simply exit, because 20<21. In order to get it to do what we want, we need to run it like this:
```
retinanet-train --batch-size=1 --steps=50 --epochs=40 --initial-epoch=21 --snapshot /snapshots/resnet50_coco_best_v2.1.0.h5 --compute-val-loss --tensorboard-dir /snapshots/logs/$(date +%m-%d-%H%M) --snapshot-path /snapshots coco /data/COCO/
```

So, `epochs` has the meaning of `total number of epochs elapsed since model training began`, rather than `number of epochs to train` during this specific training run. To me, it's a bit confusing. If you agree, I could try to modify the code so that  

```
retinanet-train --batch-size=1 --steps=50 --epochs=20 --initial-epoch=21 --snapshot /snapshots/resnet50_coco_best_v2.1.0.h5 --compute-val-loss --tensorboard-dir /snapshots/logs/$(date +%m-%d-%H%M) --snapshot-path /snapshots coco /data/COCO/
```

does what I would expect. However, if you prefer the current API behavior, I won't touch this.

bfialkoff:
The current behavior is designed to match the behavior of keras. I think it
inadvisable to change it.

hgaiser:
I don't feel like there is one clear expected behavior, comparing the current with the proposed by @AndreaPi . Personally I would expect the current behavior, because the argument is called `--epochs`. Had it been called `--nr-epochs` I would have expected @AndreaPi 's proposed behavior. I think we should keep it like this for now since it indeed also matches with what Keras does.

AndreaPi:
Good points, guys. I agree with you.

PR #1162, opened 2019-11-01
Added inference time measure.

SalahAdDin:
This kind of measure are so important in order to compare this thecnique with others.

PR #1183, opened 2019-11-15
Callback Evaluate expect tuple returned by utils.eval.evaluate 

anilkens:
A call to `utils.eval.evaluate` in `callbacks.eval.Evaluate` didn't reflect that the inference time is returned next to the dict of precisions.
(I didn't know whether there's a preferred way of ignoring return values within the code base, so this fix is just an arbitrary suggestion.)

### DIFF
--- a/keras_retinanet/callbacks/eval.py
+++ b/keras_retinanet/callbacks/eval.py
@@ -60,7 +60,7 @@ class Evaluate(keras.callbacks.Callback):
         logs = logs or {}
 
         # run evaluation
-        average_precisions = evaluate(
+        average_precisions, inference_time = evaluate(
             self.generator,
             self.model,
             iou_threshold=self.iou_threshold,


de-vri-es:
It's common in python to use `_` for ignoring tuple elements. However, it's also kinda nice to see what is being ignored. Prefixing the name with an underscore (so `_inference_time`) is common in that case. Personally I'm also fine with this.

However, I'm slightly concerned that this may not be entirely backwards compatible. Why did this work in the past? Or did it never work? (Yay Python and lack of static type checking.) Do you know if and when this was changed in Keras?

hgaiser:
Ah yes, this was introduced in https://github.com/fizyr/keras-retinanet/pull/1162 . I agree with @de-vri-es regarding the naming convention though.

de-vri-es:
Oh right, it's an internal thing. Then nevermind about backwards compatibility :)

hgaiser:
Thanks again for the fix :+1:

anilkens:
Hi, sorry, I missed the discussion, but totally agree that a leading underscore makes a nice naming convention.
Thanks for the quick merging and for the nice repository in the first place. :)

PR #1192, opened 2019-11-26
modified sequence of densenet

Hannah37:
I've tried to use densenet as a backbone but the model detected senet instead of densenet. 
Since 'senet' is a substring of 'densenet', the conditional statement below (models/__init__.py) could not reach to densenet.
```
def backbone(backbone_name):
    """ Returns a backbone object for the given backbone.
    """
    if 'seresnext' in backbone_name or 'seresnet' in backbone_name or 'senet' in backbone_name:
        from .senet import SeBackbone as b
    elif 'resnet' in backbone_name:
        from .resnet import ResNetBackbone as b
    elif 'mobilenet' in backbone_name:
        from .mobilenet import MobileNetBackbone as b
    elif 'vgg' in backbone_name:
        from .vgg import VGGBackbone as b
    elif 'densenet' in backbone_name:
        from .densenet import DenseNetBackbone as b
    elif 'EfficientNet' in backbone_name:
        from .effnet import EfficientNetBackbone as b
    else:
        raise NotImplementedError('Backbone class for  \'{}\' not implemented.'.format(backbone))

    return b(backbone_name)
```

### DIFF
--- a/keras_retinanet/models/__init__.py
+++ b/keras_retinanet/models/__init__.py
@@ -49,16 +49,16 @@ class Backbone(object):
 def backbone(backbone_name):
     """ Returns a backbone object for the given backbone.
     """
-    if 'seresnext' in backbone_name or 'seresnet' in backbone_name or 'senet' in backbone_name:
+    if 'densenet' in backbone_name:
+        from .densenet import DenseNetBackbone as b
+    elif 'seresnext' in backbone_name or 'seresnet' in backbone_name or 'senet' in backbone_name:
         from .senet import SeBackbone as b
     elif 'resnet' in backbone_name:
         from .resnet import ResNetBackbone as b
     elif 'mobilenet' in backbone_name:
         from .mobilenet import MobileNetBackbone as b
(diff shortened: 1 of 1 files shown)


hgaiser:
Aii I see what you mean. Could you make a PR to fix this? I think for now just moving the densenet check up would be fine, so that it get's checked before senet.

Hannah37:
> Aii I see what you mean. Could you make a PR to fix this? I think for now just moving the densenet check up would be fine, so that it get's checked before senet.

This is the PR i sent with the comment above. I just moved the densenet up and hope it is merged asap.

hgaiser:
> This is the PR i sent with the comment above. I just moved the densenet up and hope it is merged asap.

Haha I'm an idiot, I see now :)



> It seems travis-ci has some problems while building...but i have no idea of the reason it failed. The only change i made is just moving the densenet up. This the first time i sent a PR, so let me know if I did something wrong.

Yep, there's some issue with Keras / Tensorflow. We can ignore that for now.

PR #1204, opened 2019-12-02
Speeding up the pre-processing

vsuryamurthy:
As observed from our experiments, this change speeds up the pre-processing upto 1 second, depending on the size of the input image.


hgaiser:
LGTM

Issue #1219, opened 2019-12-13
Syntax error (missing a comma) at line 522 in train.py


sv-youn:
Line 522 is missing a comma at the end of the statement

de-vri-es:
Thanks, PR opened in #1221

de-vri-es:
Actually, #1220 was first.

PR #1221, opened 2019-12-13
Fix syntax error in train.py

de-vri-es:
This PR fixes a syntax error in `train.py`, reported in #1219 by @sv-youn.


de-vri-es:
Woops, already fixed in #1220

PR #1252, opened 2020-01-21
Update README.md

fnobis:
Add link to a camera and radar fusion network that is built from a baseline retina network


hgaiser:
Thank you, looks very cool!

PR #1260, opened 2020-01-30
Adding the image source

SalahAdDin:
it makes more easy to know which the annotation that is giving problems.


hgaiser:
Thanks :+1:

Issue #1259, opened 2020-01-30
Support PNG extension to PASCAL VOC


SalahAdDin:
Hi,

I'm working with the **ACFR** dataset and i got a problem: its images are all in **PNG** format, and as far as i can see this RetinaNet implementation use JPEG images.

So, can we add support to PNG or multi-format image? Thanks.

hgaiser:
Yes that would be nice to have. Could you submit a PR to add functionality to change what extension is to be expected for Pascal VOC?

SalahAdDin:
@hgaiser  Should we do it automatically or manually?

hgaiser:
Unfortunately the filenames in the .txt files don't include an extension, otherwise we could just use that ... I would choose for manually, with the default being whatever it is that Pascal VOC has (jpg I believe).

SalahAdDin:
The other way is "walk" along the PASCAL VOC folder and read each file in the folder and pass after as the image, but it could lessen the performance.

de-vri-es:
If the PASCAL VOC dataset has a folder per image, a searching the folder for the right file isn't such a weird idea. Overhead is probably negligible, and it would support datasets with mixed file extensions.

It would even solve incompatibilities with Windows if people wrongly capitalize file extensions.

SalahAdDin:
@de-vri-es  You got a good point: i don't know it is offend possible to find different file's extensions.

@hgaiser What do you think?

SalahAdDin:
@de-vri-es 
There is a problem with walking through the directory: https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/preprocessing/pascal_voc.py#L94
```
self.image_names          = [l.strip().split(None, 1)[0] for l in open(os.path.join(data_dir, 'ImageSets', 'Main', set_name + '.txt')).readlines()]
 ```

It is taking the files from the training, testing or validation files, and it is logical because it will be used in different `bin` files; that's why it does not use a walking algorithm.

In this case, i think the best idea is to put the extension manually, while the researchers must be careful about their images' extension.

PR #1262, opened 2020-02-01
Adding a parameter to handle the dataset images' extensions

SalahAdDin:
The aim is to solve this issue: https://github.com/fizyr/keras-retinanet/issues/1259


keras_retinanet/bin/debug.py
```diff
@@ -181,6 +182,7 @@ def csv_list(string):
     parser.add_argument('--random-transform', help='Randomly transform image and annotations.', action='store_true')
     parser.add_argument('--image-min-side', help='Rescale the image so the smallest side is min_side.', type=int, default=800)
     parser.add_argument('--image-max-side', help='Rescale the image if the largest side is larger than max_side.', type=int, default=1333)
+    parser.add_argument('--image-extension',   help='Declares the dataset images\' extension.', default='.jpg')
```
hgaiser: This argument is only necessary for Pascal VOC, so we should add it to that subparser.

SalahAdDin: In all `bin` script files, right?

keras_retinanet/preprocessing/pascal_voc.py
```diff
@@ -77,7 +77,7 @@ def __init__(
         data_dir,
         set_name,
         classes=voc_classes,
-        image_extension='.jpg',
+        image_extension,
```
hgaiser: It's not allowed to have a positional argument after optional arguments.

SalahAdDin: Image extension must be before to classes, right?

hgaiser:
```python
def foo(a, b=1, c):
	print(a, b, c)

foo(1, 2)
```

```shell
  File "test.py", line 1
    def foo(a, b=1, c):
            ^
SyntaxError: non-default argument follows default argument
```

You can give the `image_extension` a default value, making it an optional argument.

SalahAdDin: Ok,

keras_retinanet/bin/debug.py
```diff
@@ -94,6 +94,7 @@ def create_generator(args):
         generator = PascalVocGenerator(
             args.pascal_path,
             args.pascal_set,
+            args.image_extension,
```
hgaiser: Without the argument name this is assigned to `classes`.

SalahAdDin: I don't get you.

hgaiser: The arguments for the `PascalVocGenerator` are:
```python
    def __init__(
        self,
        data_dir,
        set_name,
        classes=voc_classes,
        image_extension='.jpg',
        skip_truncated=False,
        skip_difficult=False,
        **kwargs
    ):
```

So the third argument would be `classes`, where you pass `args.image_extension`.

keras_retinanet/bin/evaluate.py
```diff
@@ -54,6 +54,7 @@ def create_generator(args):
         validation_generator = PascalVocGenerator(
             args.pascal_path,
             'test',
+            args.image_extension,
```
hgaiser: Same issue here.

keras_retinanet/bin/train.py
```diff
@@ -266,15 +266,17 @@ def create_generators(args, preprocess_image):
     elif args.dataset_type == 'pascal':
         train_generator = PascalVocGenerator(
             args.pascal_path,
-            'trainval',
+            'train',
+            args.image_extension,
```
hgaiser: And here.

keras_retinanet/bin/train.py
```diff
@@ -266,15 +266,17 @@ def create_generators(args, preprocess_image):
...
             transform_generator=transform_generator,
             visual_effect_generator=visual_effect_generator,
             **common_args
         )
 
         validation_generator = PascalVocGenerator(
             args.pascal_path,
-            'test',
+            'val',
+            args.image_extension,
```
hgaiser: And here.

hgaiser:
I see this also includes the move from `trainval` to `train` and from `test` to `val`. I checked Pascal VOC 2007 and it seems fine to me.

Thanks for your help!

SalahAdDin:
Your welcome, Now, you can see the F1 score, it works now.

Issue #930, opened 2019-03-05
Error


HuangQinJian:
keras_retinanet/bin/../../keras_retinanet/preprocessing/generator.py:170: UserWarning: Image with id 244 (shape (466, 542, 3)) contains the following invalid boxes: [[430. 183. 595. 285.]].
  annotations['bboxes'][invalid_indices, :]
keras_retinanet/bin/../../keras_retinanet/preprocessing/generator.py:170: UserWarning: Image with id 1033 (shape (465, 1083, 3)) contains the following invalid boxes: [[451. 444. 518. 489.]
 [454. 477. 519. 561.]].
  annotations['bboxes'][invalid_indices, :]
keras_retinanet/bin/../../keras_retinanet/preprocessing/generator.py:170: UserWarning: Image with id 896 (shape (400, 527, 3)) contains the following invalid boxes: [[495. 426. 622. 552.]].
  annotations['bboxes'][invalid_indices, :]
keras_retinanet/bin/../../keras_retinanet/preprocessing/generator.py:170: UserWarning: Image with id 1100 (shape (527, 520, 3)) contains the following invalid boxes: [[406. 344. 634. 426.]
 [523. 405. 594. 535.]
 [105. 418. 197. 539.]
 [345. 470. 522. 507.]].

What does it mean?

hgaiser:
It means your dataset contains bounding boxes that lie outside the image. Your first warning for instance has `x2 = 595` but your width of your image is `542`.

Since this is not an issue with our code, I'll close this. Please continue the discussion on Slack.

HuangQinJian:
I have a question.The origin image size was   (466, 542, 3), but I want to scale it to(800,600,3),so I aslo rescaled the box?It is wrong?Or the input size can be random?Or where can I set the input_size?

hgaiser:
You can set the [`image-min-side` and `image-max-side`](https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/bin/train.py#L409-L410) arguments in train.py. This will automatically scale your images and bboxes accordingly.

HuangQinJian:
So ,I do not need to sacle box in annotations.csv?

SalahAdDin:
@hgaiser Is there a way to know which image is the problematic one? I tried looking by the boxes but i didn't find any box in my dataset matched with.

hgaiser:
If you have the generator, you can get the image path using `generator.image_path(image_index)`.

[...]

PR #1263, opened 2020-02-01
Anchors fixes: reduce the limits increasing x and y values

SalahAdDin:
This is related to this issue: https://github.com/fizyr/keras-retinanet/issues/930

I think the logic is to reduce the bounding box area in order to make them be lower than the image size, but i'm not sure about that.

We can discuss about this.

Merge this before: https://github.com/fizyr/keras-retinanet/pull/1262


hgaiser:
Why would you clip to `width + 1`?

SalahAdDin:
I'm not sure about it.

PR #1270, opened 2020-02-10
Tensorflow2 compatibility

hgaiser:
Attempt number twentyfive.

With a bit of luck this fixes the unit tests, let's hold off merging until we see the result :)

ps. we should decide if we want the minimum tf version to be 2.1.0 or not. 1.14 and 1.15 should also be fine, but 2.0.0 is broken..


vcarpani:
LGTM

Issue #1055, opened 2019-07-03
predict scores is lower than evaluate score


kolyadin:
Hi!

I have successfully train and evaluate model.
My evaluate stats:

> Running network: 100% (257 of 257) |######################################################################################################| Elapsed Time: 0:17:19 Time:  0:17:19
> Parsing annotations: 100% (257 of 257) |##################################################################################################| Elapsed Time: 0:00:00 Time:  0:00:00
> 0 instances of class nothing with average precision: 0.0000
> 257 instances of class cashier_check with average precision: 1.0000
> mAP using the weighted average of precisions among classes: 1.0000

So precisions is "1.0000" (i am evaluate on same train data)

But when i run predict from this example:
https://github.com/fizyr/keras-retinanet/blob/master/examples/ResNet50RetinaNet.ipynb

I receive smaller precisions (e.g. "score": 0.905562162399292) on same images.

Why predict scores differs from evaluate scores?
Thanks.

jsemric:
Hi.

Maybe you filtered out the predictions with a low score (below 0.5). In the evaluation, the score threshold is 0.05 by default. Also, make sure you feed the images in the BGR format to the model.

hgaiser:
@jsemric has a good point, the default score threshold for evaluation is 0.05, for the example notebook it is set to 0.5. This could explain the difference you're seeing.

hgaiser:
Any update here?

ikerodl96:
Hello,
Today I have experienced the same issue. In my case the problem was that the data generator in _evaluate.py_ was not created by using the specific data preprocessing method associated to the selected model backbone. Instead, the default _caffe_ method was used. To solve this, I have relied on the code that implements both the train and validation data generators in _[train.py](https://github.com/fizyr/keras-retinanet/blob/e27df4c3c8f003dd1c3e0f2e68adbfcfd2366213/keras_retinanet/bin/train.py#L211)_, as in this case, all the additional parameters that influence the generators are considered (the particular backbone preprocessing method is passed as a parameter to the _create\_generators()_ function and other arguments such as the batch size or the image min side are stored in the _common\_args_ dict). 
Hope it helps and that you can solve the problem.

hgaiser:
Thanks for letting us know @ikerodl96 , could be related to #647 . I'm assuming the original issue is resolved though.

mariaculman18:
Hi @ikerodl96, @hgaiser,

[...]

Issue #647, opened 2018-08-26
Evaluation always produces mAP of 0.0 when using backbones other than Resnet50


jpxrc:
First and foremost, thank you for the awesome package! The dataset I am using is of satellite images consisting of 29 different classes. I have been able to train and evaluate a retinanet model on this dataset using the default 'densenet50' backbone on a subset of the 29 classes. 

However, when I switch over to training and evaluating a model with a different backbone network such as 'densenet121', all of the mAP scores for each class is zero. I'm not receiving any issues when training (I am also using the random-transform flag for each epoch), or when converting the model (I also supply the --backbone='densenet121' flag) and it converts successfully. I can also see that losses being optimized during training so it's definitely detecting and classifying the objects in the images.

I even tried using the original resnet50 model trained on a subset of classes to see if it would pick up those classes on the full dataset with 29 classes and it still produces an output of zero. I looked at the validation_annotations.csv file for both cases and the formatting is identical so I don't think it has to do with the annotation files.

I have attached the validation_annotations.csv file, the classes.csv file (converted to .txt files in order to attach them here)
[common_classes.txt](https://github.com/fizyr/keras-retinanet/files/2321190/common_classes.txt)
[common_validation_annotations.txt](https://github.com/fizyr/keras-retinanet/files/2321191/common_validation_annotations.txt)

Any ideas what could be going on?

EDIT: I just did a comparison of a Resnet50 model and Densenet121 model both trained on the same dataset that I know for sure works and the problem is definetely with the densenet121 implementation because the Resnet50 model is producing output during evaluation.

yyannikb:
I have the same problem for densenet121. When I read out the dense classification scores (for all anchors) they are much too confident (most are very close to or equal to 1).

hgaiser:
Densenet is a community contribution and I have never really used it. If you find out what the problem is then a PR will be welcome :)

songhuizhong:
I have the same problem when i changed the backbone to the shuffleNet. The loss is decreasing, but the MAP is always zero. However when i trained model with resnet50 backbone, everything is okay. I still did not find the problem. Can anyone give me some advice?

songhuizhong:
yeah, I evaluate the map during training. The dataset i used is pascal voc2007. And I did not use pretrained model, because i did not find the  pretrained weights of shufflenet-V2 on imagenet.

hgaiser:
[...]

PR #1290, opened 2020-02-25
Pass preprocess_image function to generators in evaluate.py

prickly-u:
I've encountered a problem similar to this issue: https://github.com/fizyr/keras-retinanet/issues/647 (with custom backbone, mobilenetv3, not in the "zoo"). Here https://github.com/fizyr/keras-retinanet/issues/1055#issuecomment-523873503 @mariaculman18 offered a solution that worked for me and seems to be reasonable. The author is not going to create a pull request, but I decided to do so, as other users with non-resnet50 backbones can be affected.

### DIFF
--- a/keras_retinanet/bin/evaluate.py
+++ b/keras_retinanet/bin/evaluate.py
@@ -28,6 +28,7 @@ if __name__ == "__main__" and __package__ is None:
 from .. import models
 from ..preprocessing.csv_generator import CSVGenerator
 from ..preprocessing.pascal_voc import PascalVocGenerator
+from ..utils.anchors import make_shapes_callback
 from ..utils.config import read_config_file, parse_anchor_parameters
 from ..utils.eval import evaluate
 from ..utils.gpu import setup_gpu
@@ -35,9 +36,13 @@ from ..utils.keras_version import check_keras_version
 from ..utils.tf_version import check_tf_version
 
 
-def create_generator(args):
+def create_generator(args, preprocess_image):
     """ Create generators for evaluation.
     """
+    common_args = {
+        'preprocess_image': preprocess_image,
+    }
+
     if args.dataset_type == 'coco':
         # import here to prevent unnecessary dependency on cocoapi
         from ..preprocessing.coco import CocoGenerator
@@ -49,6 +54,7 @@ def create_generator(args):
             image_max_side=args.image_max_side,
             config=args.config,
             shuffle_groups=False,
+            **common_args
         )
     elif args.dataset_type == 'pascal':
         validation_generator = PascalVocGenerator(
@@ -59,6 +65,7 @@ def create_generator(args):
             image_max_side=args.image_max_side,
             config=args.config,
             shuffle_groups=False,
+            **common_args
         )
     elif args.dataset_type == 'csv':
         validation_generator = CSVGenerator(
@@ -68,6 +75,7 @@ def create_generator(args):
             image_max_side=args.image_max_side,
             config=args.config,
             shuffle_groups=False,
+            **common_args
         )
     else:
         raise ValueError('Invalid data type received: {}'.format(args.dataset_type))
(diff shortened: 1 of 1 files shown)


keras_retinanet/bin/evaluate.py
```diff
@@ -141,6 +150,7 @@ def main(args=None):
     # load the model
     print('Loading model, this may take a second...')
     model = models.load_model(args.model, backbone_name=args.backbone)
+    generator.compute_shapes = make_shapes_callback(model)
```
hgaiser: This can also be passed to the constructor of generator. It does mean that the order of creating generators followed by loading of the model has to be the other way around, but I don't think this is an issue.

hgaiser:
Thank you for the PR! I haven't gotten around to do this so it is much appreciated. Does the callback during training work properly? I could imagine it needs a similar treatment.

prickly-u:
@hgaiser 
> Thank you for the PR! I haven't gotten around to do this so it is much appreciated. Does the callback during training work properly? I could imagine it needs a similar treatment.

Do you mean the make_shapes_callback and that it should be passed directly to generators' constructors for all models except resnet50, instead of this:

https://github.com/fizyr/keras-retinanet/blob/e27df4c3c8f003dd1c3e0f2e68adbfcfd2366213/keras_retinanet/bin/train.py#L494-L498
?

I'm not getting high mAP with my backbone during training, so it can indeed be an issue. But not for sure, as I'm just started and there can be other problems.

hgaiser:
Ah it looks like `preprocess_image` is passed correctly. Are you getting different results with the evaluation callback versus running evaluate.py?

prickly-u:
> Ah it looks like `preprocess_image` is passed correctly. Are you getting different results with the evaluation callback versus running evaluate.py?

Yes I do, when running without the fix. I'm getting 20-30-40% mAP in the evaluation of the first epochs during training, but in the evaluate.py restored snapshots always give 0.

It looks reasonable to apply the same image preprocessing during evaluation as in the training process, however I can't get  why resnet50 works without it. May be the 'tf' preprocessing mode for mobilenet backbone instead of 'caffe' for resnet causes that?

hgaiser:
> It looks reasonable to apply the same image preprocessing during evaluation as in the training process

Oh yeah definitely, was just an oversight that this wasn't handled properly.

> I can't get why resnet50 works without it.

It's because the default value follows resnet50 style preprocessing.

I was asking because I thought you meant that, even with this fix, you get different results whether you evaluate during training or evaluate after training. I get now that this isn't the case for you. Then everything looks fine to me, thanks again for the PR.

prickly-u:
@hgaiser 
Could you please also explain about `make_shapes_callback` and code snippet from train.py I've mentioned above?
`if 'vgg' in args.backbone or 'densenet' in args.backbone: `

Should there be vgg and densenet only or other backbones as well, except the resnet50?

hgaiser:
The shapes of the feature pyramid for resnet50 and probably some others are well defined, ie. you can predict how they will end up. For others, like vgg and densenet they were more difficult to predict so the decision was made to write a function that, given a model, extracts the sizes of the feature pyramid. I recall this from memory, so I could be wrong on some details.

prickly-u:
> The shapes of the feature pyramid for resnet50 and probably some others are well defined, ie. you can predict how they will end up. For others, like vgg and densenet they were more difficult to predict so the decision was made to write a function that, given a model, extracts the sizes of the feature pyramid. I recall this from memory, so I could be wrong on some details.

Wouldn't it be better to change the line:
`if 'vgg' in args.backbone or 'densenet' in args.backbone:`
with
`if not args.backbone.startswith('resnet50):`
?

As I see, there is only one `guess_shape` function. Is it designed specially for resnet50?
https://github.com/fizyr/keras-retinanet/blob/a81b313eedbfd2ab849c3aecd6b7bef1f6cf9294/keras_retinanet/utils/anchors.py#L186-L198

hgaiser:
I'm pretty sure other backbones (like mobilenet and efficientnet) work similarly as resnet50, so you would get some list of backbones there either way.

prickly-u:
> I'm pretty sure other backbones (like mobilenet and efficientnet) work similarly as resnet50, so you would get some list of backbones there either way.

Ok, thank you!

PR #1292, opened 2020-02-25
Error when using multi-gpu

sanchezismael:
If argument expected for --gpu is type int, it is not posible to train with more than two gpus


hgaiser:
Fair point :)

I think this argument is used in more scripts, could you check them?

wheemyungshin:
It still is causing probelm:
`tf.config.experimental.set_visible_devices(gpus[int(gpu_id)], 'GPU')`
as it cannot convert from '0,1,2,3' to int.

Also even if I specify some numbers like,
`tf.config.experimental.set_visible_devices(gpus[0:2], 'GPU')`
It outputs another error which is related to this page: https://github.com/keras-team/keras/issues/13684

PR #1294, opened 2020-02-27
added additional documentation aboout bbox regression targets

AldusLorde:
Sorry for a late response and PR you asked me to do - I had lots of things to do, so I forgot about it...


hgaiser:
Thanks :+1:

PR #1313, opened 2020-03-25
fix bug when setting up gpu=0

edwardyangxin:
when setting up gpu = 0 , it won't go into "if args.gpu" block


hgaiser:
Thanks :+1:

PR #1321, opened 2020-04-03
Add options to convert_model function

Svan70:
I needed to set score_threshold and max_detections for the inference model.
Given the current code It is straightforward and could be useful, that's why I 
submit this PR. It just add the possibility to set these parameters in convert_model function (and in convert_model script to be consistent).

By the way, many thanks for this package.

### DIFF
--- a/keras_retinanet/bin/convert_model.py
+++ b/keras_retinanet/bin/convert_model.py
@@ -78,14 +78,16 @@ def main(args=None):
     models.check_training_model(model)
 
     # convert the model
-    model = models.convert_model(model,
-                                 nms=args.nms,
-                                 class_specific_filter=args.class_specific_filter,
-                                 anchor_params=anchor_parameters,
-                                 nms_threshold=args.nms_threshold,
-                                 score_threshold=args.score_threshold,
-                                 max_detections=args.max_detections,
-                                 parallel_iterations=args.parallel_iterations)
+    model = models.convert_model(
+        model,
+        nms=args.nms,
+        class_specific_filter=args.class_specific_filter,
+        anchor_params=anchor_parameters,
+        nms_threshold=args.nms_threshold,
+        score_threshold=args.score_threshold,
+        max_detections=args.max_detections,
+        parallel_iterations=args.parallel_iterations
+    )
 
     # save model
     model.save(args.model_out)
--- a/keras_retinanet/bin/convert_model.py
+++ b/keras_retinanet/bin/convert_model.py
@@ -43,6 +43,10 @@ def parse_args(args):
     parser.add_argument('--no-nms', help='Disables non maximum suppression.', dest='nms', action='store_false')
     parser.add_argument('--no-class-specific-filter', help='Disables class specific filtering.', dest='class_specific_filter', action='store_false')
     parser.add_argument('--config', help='Path to a configuration parameters .ini file.')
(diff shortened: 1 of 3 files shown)


keras_retinanet/bin/convert_model.py
```diff
@@ -43,6 +43,10 @@ def parse_args(args):
     parser.add_argument('--no-nms', help='Disables non maximum suppression.', dest='nms', action='store_false')
     parser.add_argument('--no-class-specific-filter', help='Disables class specific filtering.', dest='class_specific_filter', action='store_false')
     parser.add_argument('--config', help='Path to a configuration parameters .ini file.')
+    parser.add_argument('--nms-threshold', help='Value for non maximum suppression threshold', dest='nms_threshold', type=float, default=0.5)
```
hgaiser, line 46: No need to define `dest`, these are already the defaults.

hgaiser: Bit nitpicky, but could you add a period after the `help` sentence?

keras_retinanet/bin/convert_model.py
```diff
@@ -74,7 +78,9 @@ def main(args=None):
     models.check_training_model(model)
 
     # convert the model
-    model = models.convert_model(model, nms=args.nms, class_specific_filter=args.class_specific_filter, anchor_params=anchor_parameters)
+    model = models.convert_model(model, nms=args.nms, class_specific_filter=args.class_specific_filter, anchor_params=anchor_parameters,
```
hgaiser, line 81: Also nitpicky, but our style for long lines is:
```python
foo(
    bar,
    baz=1,
    foobar=2
)
```
It's more lines, but IMO more readable.

keras_retinanet/models/__init__.py
```diff
@@ -87,14 +87,15 @@ def load_model(filepath, backbone_name='resnet50'):
...
-def convert_model(model, nms=True, class_specific_filter=True, anchor_params=None):
+def convert_model(model, nms=True, class_specific_filter=True, anchor_params=None, **kwargs):
     """ Converts a training model to an inference model.
 
     Args
         model                 : A retinanet training model.
         nms                   : Boolean, whether to add NMS filtering to the converted model.
         class_specific_filter : Whether to use class specific filtering or filter for the best scoring class only.
         anchor_params         : Anchor parameters object. If omitted, default values are used.
+        **kwargs              : Inference and minimal retinanet model settings
```
hgaiser, line 98: Similarly here, period at the end.

keras_retinanet/bin/convert_model.py
Svan70: Done

keras_retinanet/bin/convert_model.py
Svan70: Done

keras_retinanet/models/__init__.py
Svan70: Done

Svan70:
Are there any unitary / integrated tests I have to run / write ?

keras_retinanet/bin/convert_model.py
hgaiser: Not exactly what I had in mind, could you have another look at this?

Svan70: Ok. I did it similarily to your other function calls. Like this ?

hgaiser:
> Are there any unitary / integrated tests I have to run / write ?

Normally I'd say yes, but in this case it is just argument passing. Thanks again for the PR!

PR #1334, opened 2020-04-18
correct description of anchor_targets_bbox() function parameters

Gateway2745:
annotations_group is a list of dictionaries which was incorrectly described as "np.array of shape (N, 5) for (x1, y1, x2, y2, label)" previously


hgaiser:
Yes, thank you!

PR #1352, opened 2020-05-10
earlystoppping map

brunoeducsantos:
Hi,
The goal is creating an early stopping as function of mean average precision metric.
Regards,
Bruno


hgaiser:
Looks good to me, thanks :)

Issue #1349, opened 2020-05-08
Label Smoothing


waylonflinn:
Suggestions on how to implement label smoothing?

It looks like a modification in `anchor_targets_bbox` may be necessary, but what kind? It looks like this produced a label for each anchor but I don't really understand the format of the labels produced or how they relate to the input classes. Can you speak a bit on that or point me to documentation or other explanation?

Thanks.

hgaiser:
It seems you can already use label smoothing from Keras:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/losses.py#L558-L561

Which we use here:

https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/losses.py#L58

waylonflinn:
Thanks for the information. I don't currently understand how binary crossentropy is used to handle more than two classes, but this may work.

pleaseRedo:
How can we integrate `keras.losses.binary_crossentropy`  with label_smoothing enabled into our code? 
As you mentioned in 
https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/losses.py#L58
it is `keras.backend.binary_crossentropy()` but it does not have arg like `label_smoothing`
`keras.losses.binary_crossentropy()` allows `label_smoothing`, but changing `keras.backend.binary_crossentropy(labels, classification)` to `keras.losses.binary_crossentropy(labels,classification,label_smoothing=0.1)` function causes OOM issues, so I guess I'm using it in the wrong way?

waylonflinn:
A note: this functionality (`label_smoothing`) was added to keras in 2.3.0

waylonflinn:
Thanks for everyone's comments here. I've dug into both the implementation of focal loss in `keras_retinanet` and the implementation of `binary_cross_entropy` in keras and I think I understand both how the current implementation works (for hard targets) and why it won't work for soft targets (including those found in label smoothing).
I'm going to work through all the math to verify and then attempt to formulate a solution. If I manage to solve this issue are you open to a Pull Request?

waylonflinn:
I've implemented a simple fix that allows soft targets (including smoothed labels) to work with the current loss function implementation. 

Even though I still have reservations about using this loss function for multiclass classification, I'll put together a PR later today, after a bit more testing.

waylonflinn:
Here's the code I'm using to do the actual smoothing.

```python
def smooth_batch_labels(batch_labels, factor=0.1):
    """
[...]

PR #1357, opened 2020-05-17
Modify Focal Loss to Allow Soft Targets

waylonflinn:
Change the selector for alpha and focal weights to choose indices greater than a cutoff (0.5 by default) instead of equal to 1. Functionality should be identical for hard targets (one-hot encoding) and permit soft targets, such as those used in label smoothing, to function in a reasonable way.

This change allows external (user) implementation of functionality in #1349

### DIFF
--- a/keras_retinanet/losses.py
+++ b/keras_retinanet/losses.py
@@ -18,12 +18,13 @@ import keras
 from . import backend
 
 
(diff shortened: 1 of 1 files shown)


hgaiser:
Looks good to me :) thanks

PR #1370, opened 2020-05-26
parameters for learning rate reduction upon loss plateau mutable by CLI

mooratov:
the default parameters for ReduceLROnPlateau are fairly aggressive. Probably fine for most cases, but can be a problem if number of steps per epoch is low or validation set is limited.


hgaiser:
LGTM, thanks :)

PR #1381, opened 2020-06-03
Pyramid level customization

mooratov:
made pyramid levels configurable, and implemented the capability to use P2 pyramid layer. 
Currently, P2 layers only works for Resnet backbone.

### DIFF
--- a/keras_retinanet/bin/convert_model.py
+++ b/keras_retinanet/bin/convert_model.py
@@ -66,10 +66,15 @@ def main(args=None):
 
     # optionally load config parameters
     anchor_parameters = None
+    pyramid_levels = None
     if args.config:
         args.config = read_config_file(args.config)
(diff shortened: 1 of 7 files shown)


mooratov:
very sorry, this was supposed to go on my personal fork (not ready for primetime)

muaali:
@mooratov I am trying to understand what is the benefit of using P2 and C2 in retinanet. 
Also did it benefit you in some way, in terms of performance. What was the scale / size of objects that you were trying to detect. Thanks.

mooratov:
we were interested in very small bounding boxes, on the scale of 4x4 pixels
or so.

muaali:
Thanks @mooratov . I will use your commits to include C2-P2 in the model and then to train the model and see if it learns better.

PR #1387, opened 2020-06-13
Ignore files, update requirements

ppizarror:
keras-resnet v0.2.0 works good, so I've added >=0.1.0 for compatibility
Also, as my python installation was clean, I figured out that `pycocotools` library is required in tests, and I had not installed it, and it was not flagged as required in tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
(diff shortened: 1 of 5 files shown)


hgaiser:
keras-resnet 0.2.0 has a backwards incompatible change: https://github.com/broadinstitute/keras-resnet/issues/57

ppizarror:
I understand, thanks!

Issue #1366, opened 2020-05-25
RuntimeWarning: Early stopping conditioned on metric `mAP` which is not available


DmitryRyumin:
Hello everyone! Why am I receiving this warning?

RuntimeWarning: Early stopping conditioned on metric `mAP` which is not available. Available metrics are: loss,regression_loss,classification_loss,lr   (self.monitor, ','.join(list(logs.keys()))), RuntimeWarning

hgaiser:
Did you change something? I don't believe we configure early stopping on mAP.

DmitryRyumin:
No! I didn't change anything. Latest Versions TF and Keras. Can I link to this metric?

ppizarror:
Hi, mAP metric is added by Evaluation class in `keras_retinanet.utils`. Evaluation needs a validation_generator for computing the metric, so you need to define:

1. Enable evaluation on your model
2. Provide a validation dataset

From `train.py`:

```python
    if args.evaluation and validation_generator:
        if args.dataset_type == 'coco':
            from ..callbacks.coco import CocoEval

            # use prediction model for evaluation
            evaluation = CocoEval(validation_generator, tensorboard=tensorboard_callback)
        else:
            evaluation = Evaluate(validation_generator, tensorboard=tensorboard_callback,
                                  weighted_average=args.weighted_average) <--- check this
        evaluation = RedirectModel(evaluation, prediction_model)
        callbacks.append(evaluation)
```

PR #1388, opened 2020-06-13
Enable earlystopping only if validation set is provided

ppizarror:
Related issue: #1366
EarlyStopping class uses mAP by default, but mAP only is computet if evaluation dataset is provided. If one tries to train the model without any validation dataset, a warning appears:

`RuntimeWarning: Early stopping conditioned on metric mAP which is not available. Available metrics are: loss,regression_loss,classification_loss,lr (self.monitor, ','.join(list(logs.keys()))), RuntimeWarning`


hgaiser:
Awesome, thank you :)

PR #1398, opened 2020-06-25
Add LogoDet to "Projects using keras-retinanet"

bedapudi6788:


hgaiser:
Sorry for the delay, but thanks for the PR!

PR #1401, opened 2020-06-27
Remove bboxes clipping

### LINKED ISSUE #882: BBOX issues
Please make sure that you follow the steps below when creating an issue.
Only use GitHub issues for issues with the implementation, not for issues with specific datasets or general questions about functionality.
If your issue is an implementation question, please ask your question on the #keras-retinanet Slack channel instead of filing a GitHub issue.
You can find directions for the Slack channel here: https://github.com/fizyr/keras-retinanet#discussions

Thank you!

1. Check that you are up-to-date with the master branch of keras-retinanet.
2. Check that you are up-to-date with the latest version of Keras: https://github.com/keras-team/keras.
3. Check that you are up-to-date with the latest version of TensorFlow.
   The installation instructions can be found here: https://www.tensorflow.org/get_started/os_setup.
4. Check that you have read the entire README.md: https://github.com/fizyr/keras-retinanet/README.md.
   Most noticably the FAQ section shows common issues: https://github.com/fizyr/keras-retinanet#faq.
5. Clearly describe the issues you're having including the expected behaviour, the actual behaviour
   and the steps required to trigger the issue.
6. Include relevant output from the commands you're executing, including full stack traces where relevant.
7. Remove this entire message and replace it with your issue.

I have debug my training samples by myself. When train the model using my own datasets, the issue is as below:
image
The bboxes are within the image. I do not know why it shows Image contains the following invalid boxes. 
The bboxes are reasonable. Could you please give a explanation?

rafis:
fixes #882

### DIFF
--- a/keras_retinanet/layers/__init__.py
+++ b/keras_retinanet/layers/__init__.py
@@ -1,2 +1,2 @@
-from ._misc import RegressBoxes, UpsampleLike, Anchors, ClipBoxes  # noqa: F401
+from ._misc import RegressBoxes, UpsampleLike, Anchors  # noqa: F401
 from .filter_detections import FilterDetections  # noqa: F401
--- a/keras_retinanet/layers/_misc.py
+++ b/keras_retinanet/layers/_misc.py
@@ -160,26 +160,3 @@ class RegressBoxes(keras.layers.Layer):
         })
 
         return config
-
-
-class ClipBoxes(keras.layers.Layer):
-    """ Keras layer to clip box values to lie inside a given shape.
-    """
-    def call(self, inputs, **kwargs):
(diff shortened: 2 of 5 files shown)


hgaiser:
How does this fix #882 ? The error comes from loading the annotations, but you removed the clipping from the output of the network?

rafis:
Yea, it is not optimal solution. Some annotation tools does not support annotating bboxes outside image shape. But I have tested this code and it works. Model successfully predicts those bboxes as like object occluded by another object.

hgaiser:
Ah I see now that you also removed the check from the generator. I'm not a big fan of this change. In my opinion the annotations should be inside the image and so should the predictions.

PR #1437, opened 2020-08-11
Add group-method arg for train

hansoli68:
Added an argument `group-method` for `retinanet-train`.

Current default method is very slow when the dataset was large; it reads whole dataset in group_images method.


hgaiser:
Thank you for the PR! Could you add this argument to the other scripts as well? Would be useful in all scripts where we create a generator.

hansoli68:
I added args to retinanet-debug and retinanet-evaluate.

hgaiser:
:+1:

PR #1440, opened 2020-08-15
Pass optimizer-clipnorn and tensorbord update-freq as comand-line args

prickly-u:
There are a pair of arguments that can help to finetune a model when available to pass from command line:
1) Clipnorm parameter of the optimizer is hardcoded now, while it can affect training process. (It *does* at our project) 
2) Tensorboard update-freq parameter that is set to 'epoch' by default, and that can make the learning curve not so informative.

### DIFF
--- a/keras_retinanet/bin/train.py
+++ b/keras_retinanet/bin/train.py
@@ -438,7 +438,7 @@ def parse_args(args):
     parser.add_argument('--optimizer-clipnorm', help='Clipnorm parameter for  optimizer.', type=float, default=0.001)
     parser.add_argument('--snapshot-path',    help='Path to store snapshots of models during training (defaults to \'./snapshots\')', default='./snapshots')
     parser.add_argument('--tensorboard-dir',  help='Log directory for Tensorboard output', default='')  # default='./logs') => https://github.com/tensorflow/tensorflow/pull/34870
-    parser.add_argument('--tensorboard-freq', help='Update frequency for Tensorboard output. Values \'epoch\', \'batch\' or int', type=str, default='epoch')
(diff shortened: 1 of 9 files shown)


keras_retinanet/bin/train.py
```diff
@@ -430,8 +434,10 @@ def csv_list(string):
     parser.add_argument('--epochs',           help='Number of epochs to train.', type=int, default=50)
     parser.add_argument('--steps',            help='Number of steps per epoch.', type=int, default=10000)
     parser.add_argument('--lr',               help='Learning rate.', type=float, default=1e-5)
+    parser.add_argument('--optimizer-clipnorm', help='Clipnorm parameter for  optimizer.', type=float, default=0.001)
     parser.add_argument('--snapshot-path',    help='Path to store snapshots of models during training (defaults to \'./snapshots\')', default='./snapshots')
     parser.add_argument('--tensorboard-dir',  help='Log directory for Tensorboard output', default='')  # default='./logs') => https://github.com/tensorflow/tensorflow/pull/34870
+    parser.add_argument('--tensorboard-freq', help='Update frequency for Tensorboard output. Values \'epoch\', \'batch\' or int', type=str, default='epoch')
```
hgaiser, line 440: `str` is the default type, no need to provide it.

hgaiser (commented):
Thanks for the PR! One small comment, and could you rebase on master?

keras_retinanet/bin/train.py
prickly-u: > Thanks for the PR! One small comment, and could you rebase on master?
Well, I've made the correction and rebased the branch, that lead to pulling a number of other commits. Was it done right?

hgaiser:
Thank you :+1:

Issue #1444, opened 2020-08-27
retinanet-train gives error


mavalderrama:
Hello dev team, thanks for this awesome work.

I was working with retinanet last 2 weeks and everything was ok since today when I built a new container with a fresh copy of the repository.

I'm getting this error below.

Thanks for your help.

2020-08-27 23:01:26.378718: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1247] Created TensorFlow device (/job:localhost/replica:0/task:0/device:GPU:0 with 28454 MB memory) -> physical GPU (device: 0, name: GRID V100D-32Q, pci bus id: 0000:02:04.0, compute capability: 7.0)
2020-08-27 23:01:26.380208: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x55d3d57dd0f0 initialized for platform CUDA (this does not guarantee that XLA will be used). Devices:
2020-08-27 23:01:26.380232: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): GRID V100D-32Q, Compute Capability 7.0
Traceback (most recent call last):
  File "/opt/conda/envs/retinanet/bin/retinanet-train", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/envs/retinanet/lib/python3.6/site-packages/keras_retinanet/bin/train.py", line 516, in main
    config=args.config
  File "/opt/conda/envs/retinanet/lib/python3.6/site-packages/keras_retinanet/bin/train.py", line 121, in create_models
    prediction_model = retinanet_bbox(model=model, anchor_params=anchor_params, pyramid_levels=pyramid_levels)
  File "/opt/conda/envs/retinanet/lib/python3.6/site-packages/keras_retinanet/models/retinanet.py", line 400, in retinanet_bbox
    )([boxes, classification] + other)
  File "/opt/conda/envs/retinanet/lib/python3.6/site-packages/tensorflow/python/keras/engine/base_layer.py", line 922, in __call__
    outputs = call_fn(cast_inputs, *args, **kwargs)
  File "/opt/conda/envs/retinanet/lib/python3.6/site-packages/tensorflow/python/autograph/impl/api.py", line 265, in wrapper
    raise e.ag_error_metadata.to_exception(e)
tensorflow.python.autograph.impl.api.StagingError: in user code:

    /opt/conda/envs/retinanet/lib/python3.6/site-packages/keras_retinanet/layers/filter_detections.py:178 call  *
[...]

PR #1422, opened 2020-07-25
Add compatibility for Keras 2.4

cbodenst:

PR #1445, opened 2020-08-28
Partially migration from keras to tf.keras

hansoli68:
PR #1422 provides keras 2.4 compatibility. Keras 2.4 just redirecting API to `tf.keras`.

So we could migration into tf.keras by just replace keras with tf.keras.

 * This PR is not a replacement of [tf-retinanet](https://github.com/fizyr/tf-retinanet). Keras dependency still exists because [keras-resnet](https://github.com/broadinstitute/keras-resnet) requires keras.
 * Minimum tensorflow version is changed to 2.3.0. (2.2.0 occurs an error in issue #1444)
 * From keras 2.4.0, I couldn't access `mobilenet.BASE_WEIGHT_PATH` so I hard-coded it.


hgaiser:
Looks good, I think we should merge this PR. I'm not so sure if this works though. As you said, keras-resnet requires keras. In my previous experience, mixing tf.keras and keras does not work out well. Did this change in keras 2.4 since it is just forwarding the API?

hansoli68:
When I tested with latest version it works. At least everything works fine since Keras 2.4 and Tensorflow 2.3.
I don't know about previous versions well; but I think so like you said. It must works even if mixed because both are the same API.

hgaiser:
Just tried it out and everything seems to work as expected, thanks!

PR #1448, opened 2020-08-30
Update effnet.py

karthikayan4u:
I guess as keras has been updated with tensorflow, efficientnet.keras is no longer available and hence I am updating this line.

### DIFF
--- a/keras_retinanet/models/effnet.py
+++ b/keras_retinanet/models/effnet.py
@@ -19,7 +19,7 @@ from keras.utils import get_file
 
 from . import retinanet
 from . import Backbone
-import efficientnet.keras as efn
+import tensorflow.keras.applications.efficientnet as efn
 
 
 class EfficientNetBackbone(Backbone):


hansoli68:
I think `effnet.py` used [efficientnet pip package](https://pypi.org/project/efficientnet/), neither core of keras or tensorflow.
Is the implementation of tf.keras compatible?

karthikayan4u:
> I think `effnet.py` used [efficientnet pip package](https://pypi.org/project/efficientnet/), neither core of keras or tensorflow.
> Is the implementation of tf.keras compatible?

Actually there is a problem. We can't use pretrained weights as tf comes with weights itself and I don't know how to integrate the imagenet weights. Rest of all is fine.

hansoli68:
@karthikayan4u 

I hard-coded imagenet weights path from tf.keras core. (see https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/applications/efficientnet.py)

Could you please review my commit and include it into this pull request? (karthikayan4u#1)

hgaiser:
What issue are you having? I'm guessing it is related to https://github.com/qubvel/efficientnet/issues/127 ? This should be fixed with https://github.com/qubvel/efficientnet/pull/128 . When that gets released, efficientnet should work with the latest tensorflow again.

karthikayan4u:
Yeah the same was the issue along with another..I couldn't use the pretrained iamgenet weights too!

gosha20777:
I use the qubvel/efficientnet package and personally know the author of this package. This package works nicely and is well compatible with the latest TF versions. 
I believe there is no need to replace this package. Plus it breaks compatibility.

gosha20777:
In addition, I think it would be nice to replace the original resnet with the resnet from this package. This can solve many problems and will allow you to abandon the keras code and completely move to TF.

PR #1466, opened 2020-09-28
Update tensorboard writer for tf version >= 2.0.0

sedthh:
- updated writer for eval.py
- updated writer for coco.py

`log_dir` is accessed via `self.tensorboard.log_dir` which can be assigned via `--tensorboard-dir` 

However, the rest of the arguments (such as `--tensorboard-freq`) are ignored.

### DIFF
--- a/keras_retinanet/callbacks/coco.py
+++ b/keras_retinanet/callbacks/coco.py
@@ -16,7 +16,6 @@ limitations under the License.
 
 from tensorflow import keras
 from ..utils.coco_eval import evaluate_coco
-from ..utils.tf_version import check_tf_version
 
 
 class CocoEval(keras.callbacks.Callback):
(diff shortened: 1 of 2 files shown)


hgaiser:
Since we don't support tensorflow < 2.3, could you update the PR to use tensorflow 2?

sedthh:
Just to make it clear, do you want me to remove the legacy code for tf < 2.0.0. and only have the new one included?
Should I check for versions or raise an Exception if the version is lower?

hgaiser:
Yeah exactly. We are already checking for tensorflow version: https://github.com/fizyr/keras-retinanet/blob/master/keras_retinanet/bin/train.py#L471-L472

Oh btw I forgot to mention it earlier, but thanks for your PR :)

sedthh:
I have imported `check_tf_version()` also because the evaluation could be called outside of main. Is this ok or should I not include it?

sedthh:
Realized `check_tf_version()` is already called at evaluate.py so it does not need to be called again in the callbacks.

hgaiser:
Yep :) thanks again, looks good to me :+1:

PR #1468, opened 2020-10-04
Added support for HDF5 dataset and an HDF5 creation tool

madisi98:
Added retinanet-build-hdf5 entry point which allows the creation of datasets in the hdf5 format and a new option 'hdf5' to retinanet-train. This allows the dataset to be loaded in main memory the whole time and drastically reduces training times.

### DIFF
--- a/keras_retinanet/bin/build_hdf5.py
+++ b/keras_retinanet/bin/build_hdf5.py
@@ -2,7 +2,7 @@ import argparse
 
 import h5py
 import numpy as np
-from tqdm import tqdm
+from progressbar import progressbar
 
 from ..preprocessing.csv_generator import CSVGenerator
 from ..models import backbone
@@ -85,7 +85,7 @@ def main():
         bboxes_group = []
         shapes_group = []
 
-        for i in tqdm(range(generator.size()), desc=f'{split}: '):
+        for i in progressbar(range(generator.size()), prefix=f'{split}: '):
             group = [i]
             image_group = generator.load_image_group(group)
             annotations_group = generator.load_annotations_group(group)
--- a/keras_retinanet/bin/build_hdf5.py
+++ b/keras_retinanet/bin/build_hdf5.py
@@ -1,0 +1,115 @@
+import argparse
+
+import h5py
+import numpy as np
+from tqdm import tqdm
+
+from ..preprocessing.csv_generator import CSVGenerator
+from ..models import backbone
+
+
+def parse():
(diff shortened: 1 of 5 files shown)


keras_retinanet/bin/build_hdf5.py
```diff
@@ -0,0 +1,115 @@
+import argparse
+
+import h5py
+import numpy as np
+from tqdm import tqdm
```
hansoli68, line 5: https://github.com/fizyr/keras-retinanet/blob/da16864ad1b16175bedba28bce018e5b5c26462b/keras_retinanet/utils/eval.py#L26

https://github.com/fizyr/keras-retinanet/blob/da16864ad1b16175bedba28bce018e5b5c26462b/keras_retinanet/utils/coco_eval.py#L23

There are other scripts using `progressbar2` instead of `tqdm`. I think we should choose one of them.

madisi98: Agreed.
Already changed and pushed changes :)

keras_retinanet/preprocessing/hdf5_generator.py
```diff
@@ -0,0 +1,104 @@
...
+        return str(image_index)
+
+    def load_image(self, image_index):
+        return self.images[image_index]
+
+    def load_annotations(self, image_index):
+        return {'labels': self.labels[image_index],
+                'bboxes': self.bboxes[image_index]}
+
+    def compute_input_output(self, group):
```
hgaiser, line 84: You override this to remove the filtering, right? Does it have a large computational impact? I'd expect it to be minimal, in which case it would be cleaner to not override this function. Do you have a measurement for this?

hgaiser (commented):
Very nice, I've been meaning to try something like this for a while.

Do you have measurements in general for how much time is gained by using the HDF5 format?

madisi98:
I haven’t made a lot of testing yet, since I haven’t had the time for it, but I was getting roughly 6x faster epochs. I have to say that my dataset is composed of large images (around 3000x2000) and with smaller images the speedup will be less significant.

keras_retinanet/preprocessing/hdf5_generator.py
madisi98: Yes, I removed filtering because filtering happens when creating the hdf5. This process relies in the CSVGenerator class which filters the annotations already, so I considered removing that

hgaiser:
I would be interested to see the differences on a more "normal" dataset like COCO. I expect the difference will be much smaller there because AFAIK the most time spent there is from anchor target generation, not data loading.

PR #1470, opened 2020-10-04
Remove scipy dependency

hansoli68:
An older revision of the code (maybe until commit 9859954) depends `scipy`. But currently never used.


hgaiser:
Thanks again ^^

PR #1563, opened 2021-11-28
Update requirements.txt

Swati81:


hgaiser:
Why these specific versions?

PR #1574, opened 2022-05-12
Update pascal_voc.py

Vipsap:
To Extract tables annotation from page


hgaiser:
Hmm I'm not sure what this is supposed to do?

Vipsap:
I  am trying to run my model which has only one object. Table detection in
pdf

PR #1578, opened 2022-06-09
Adding support for KITTI evaluation

tuberbacon:
evaluate.py wouldn't accept kitti as argument for evaluation even though the repository contains a script with the correct functionalities (../preprocessing/kitti.py). 

changes made: import kitti.py's correct function, definition of a new  kitti parser, added call to KittiGenerator() in create_generator().


hgaiser:
Thanks :) Looks good to me.

